PR #1, opened 2016-08-18
[sailfish-components-webview] Add linkClicked signal. Contributes to JB#36016

rainemak:
All default protocol handlers are disabled by default as
navigation must be disabled within html emails.

### DIFF
--- a/components/WebView.qml
+++ b/components/WebView.qml
@@ -15,12 +15,16 @@ import Sailfish.WebView 1.0
 
 RawWebView {
     id: webview
+
+    signal linkClicked(string url)
+
     active: true
 
     onViewInitialized: {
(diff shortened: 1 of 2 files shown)


rainemak:
Maybe it's worth of adding some kind of protocol handler component (singleton) that
can be used for enabling / disabling protocol perfs. Ultimately we should implement preference API  / improve QtMozEmbeded so that prefs that we think that should be available for WebView user are easily available. Embedlite-components has EmbedPrefService that can be used for reading back preferences.

rainemak:
Adding a commit on top that tests the profile path. Now links also open e.i. browser doesn't crash upon startup.

chriadam:
LGTM

PR #2, opened 2016-09-02
[sailfish-components-webview] Move components into separate imports. Contributes to JB#36226

chriadam:


chriadam:
Still some issues.

E.g., when loading http://www.456bereastreet.com/lab/html5-input-types/ and trying the various input types, it seems that the viewable area gets "squished" when the vkb opens.

E.g., when loading http://www.html5rocks.com/en/tutorials/file/dndfiles/ the file picker etc works, but it then results in a crash, on Jolla C.  I'll attach a backtrace to JB#36158.

import/WebView.qml
```diff
@@ -12,6 +12,7 @@
 import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.WebView 1.0
+import Sailfish.WebView.Components 1.0
```
rainemak: While these are components... Some how this feels that it's given too much space. All of them are dialogs but not necessary in future. Alert, Confirm, and Prompt are JS popups (input related) and input/select handlers are also input related. Hmm, @chriadam how does something like ```Sailfish.WebView.Input``` feels to you? ```Input``` is also a bit overloaded but maybe we could come up with something else from this.

import/components/componentsplugin.cpp
```diff
@@ -0,0 +1,42 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "componentsplugin.h"
+
+#include <QtQml/QQmlEngine>
+#include <QtQml/QQmlContext>
+
+namespace SailfishOS {
```
rainemak: Maybe just Sailfish namespace.

rainemak:
I'm fine with this the import "Sailfish.WebView.Components" as well. We can still change import if we see a need.

LGTM

rainemak:
Fix for file picker and popping back crash: https://git.merproject.org/mer-core/gecko-dev/merge_requests/33

Crash was also easy to trigger with about:license and flicking it back and forth really fast so that you see only grey background. Or alternatively dragging content really slowly.

chriadam:
Updated so that now there are separate Popups + Pickers imports

chriadam:
Updated to include a variety of Popups from Sailfish Browser which weren't included originally.  The "Open in new tab" context menu option is currently unimplemented.

import/popups/ContextMenu.qml
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+            onClicked: {
+                root._hide()
+                Qt.openUrlExternally(root.linkHref)
+            }
+        }
+
+        MenuItem {
+            visible: root.isNavigable
```
rainemak: Let's add explicit tabModel property for the ContextMenu and hide "open x in a new tab" context menu options if tabModel is not present.

import/WebView.qml
```diff
@@ -0,0 +1,264 @@
...
+        visible: webview.loading
+        size: BusyIndicatorSize.Large
+    }
+
+    Timer {
+        id: helper
+        property var geolocationUrls: {{}}
+        property var contextMenu
+
+        function uniqueId() {
```
rainemak: This should be webview's uniqueId not just a random id. However, with download manager I checked quickly and it looks that viewId is no longer used.

import/WebView.qml
```diff
@@ -0,0 +1,264 @@
...
+
+        repeat: false
+        running: false
+        interval: 600 // page transition delay.
+        onTriggered: {
+            openAuthDialog(authDialogWebView, authDialogData, authDialogWinId)
+        }
+    }
+
+    Component { id: filePicker; PickerCreator {} }
```
rainemak: I'd not compile these all over here. The WebPopupHandler.js of browser was exactly for this case. Can be fixed later / separately.

import/popups/ContextMenu.qml
```diff
@@ -0,0 +1,269 @@
...
+Rectangle {
+    id: root
+
+    property string linkHref
+    property string linkTitle
+    property string linkProtocol
+    property string imageSrc
+    property string contentType
+    property PageStack pageStack
+    property var tabModel // TODO: what about this one?
```
rainemak: Let's see... at least can be used now to hide open to in a new tab kind options.

When going forward we could move both historymodel and tabmodel under these components. Let's do not rush those now.

import/WebView.qml
```diff
@@ -0,0 +1,264 @@
...
+    Component { id: singleSelectPage; SingleSelectPage {} }
+
+    Component { id: authDialog; AuthDialog {} }
+    Component { id: alertDialog; AlertDialog {} }
+    Component { id: confirmDialog; ConfirmDialog {} }
+    Component { id: locationDialog; LocationDialog {} }
+    Component { id: passwordManagerDialog; PasswordManagerDialog {} }
+    Component { id: promptDialog; PromptDialog {} }
+
+    Component { id: contextMenuComponent; ContextMenu {} }
```
rainemak: Same name as in Silica. I'd not trust that import order always holds. Maybe we should rename this one to something like ```ContextPopup``` so that named imports would not be needed now.

import/plugin.h
```diff
@@ -44,6 +47,24 @@ class RawWebView : public QuickMozView
     ~RawWebView();
 };
 
+// using custom translator so it gets properly removed from qApp when engine is deleted
+class AppTranslator: public QTranslator
```
rainemak: While this is small, we could move this to Silica C++ library. Currently every plugin has a copy.

PR #3, opened 2016-09-05
[sailfish-components-webview] Disable PageStack navigation when required. Contributes to JB#36023

chriadam:
This commit ensures that PageStack navigation is disabled when the
WebView is not panned to the x-beginning or x-end (for back-step
and forward-step, respectively).

It also prevents Page grabbing within the PageStack (if the WebView
exists within a Page within a PageStack) while pinching and panning.

Contributes to JB#36023

### DIFF
--- a/components/WebView.qml
+++ b/components/WebView.qml
@@ -19,6 +19,8 @@ RawWebView {
     signal linkClicked(string url)
 
     active: true
+    property QtObject _page
+    property QtObject _pageStack
 
     onViewInitialized: {
         webview.loadFrameScript("chrome://embedlite/content/embedhelper.js");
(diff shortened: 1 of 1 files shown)


chriadam:
Requires https://git.merproject.org/mer-core/qtmozembed/merge_requests/14 - it's not nice that this requires accessing private properties of Page and PageStack, but not sure that there's any way around that?

rainemak:
Hmm... just an idea.
```
// Silica Page
WebPageContainer  {
    // Maybe WebView should be renamed to WebPage.
    webPage: Component {
        WebPage {}
    }
}
```
This way we could build controlling to the ```WebPageContainer``` and ```WebPage``` and have explicit APIs.

chriadam:
Something like https://github.com/sailfishos/sailfish-components-webview/pull/5 ?

PR #5, opened 2016-09-14
[sailfish-components-webview] Disable PageStack navigation when required. Contributes to JB#36023

chriadam:


chriadam:
Just look at the last commit, since the others are from the other PR.

import/WebView.qml
```diff
@@ -0,0 +1,286 @@
...
+                if (parentItem.hasOwnProperty(propertyName)) {
+                    return parentItem
+                }
+                parentItem = parentItem.parent
+            }
+            return null
+        }
+
+        function setActiveInPage() {
+            if (webview.active) {
```
rainemak: There could be a ```webViewPage``` property that could be also explicitly set. Futhermore, there is no need to lookup again if ```webViewPage``` has been already found. I don't think that it is a good idea to allowing moving webview between Silica Pages. So, lookup only once.

Background: when screen is blanked we should suspend the view conditionally. As a result active changes as well. So when display is turned on again view should become active again which should not result webViewPage lookup.

import/WebView.qml
```diff
@@ -0,0 +1,286 @@
...
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.WebView 1.0
+import Sailfish.WebView.Popups 1.0
+import Sailfish.WebView.Pickers 1.0
+
+RawWebView {
+    id: webview
+
```
rainemak: Maybe ```webViewPage``` could be a property of this as well.

rainemak: Please rebase.

chriadam:
Yep, I'll do these things tomorrow (I destroyed my dev env and had to recreate it today, plus I'd put Qt5.6 onto my development device so need to reflash that one with 5.2).

import/WebViewPage.qml
```diff
@@ -0,0 +1,26 @@
...
+        State {
+            name: "webViewActive"
+            when: activeWebView != null && activeWebView != undefined && activeWebView.visible && activeWebView.active
+            PropertyChanges {
+                target: pageStack
+                _noGrabbing: activeWebView.moving || activeWebView.pinching || activeWebView.dragging
+            }
+            PropertyChanges {
+                target: webViewPage
+                backNavigation: activeWebView.atXBeginning && !activeWebView.pinching
```
chriadam: I'm not sure about the "!activeWebView.pinching" check here (and the other one above)... it makes a rather horrible grab-flicker if you pinchzoom and then pan left and then attempt to backstep via swipe gesture, when this check is included.

rainemak: Let's see how this behaves. We can iterate. I'd like to get this in asap so that we can iterate and things wouldn't be blocked.

PR #7, opened 2016-09-19
[sailfish-components-webview] Add C++ WebEngine API. Contributes to JB#36227

rainemak:


import/plugin.cpp
```diff
@@ -84,52 +43,31 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
...
+    // Disable low-precision-buffer so that background underdraw works
+    // correctly.
+    engineSettings->setPreference("layers.low-precision-buffer", QVariant::fromValue<bool>(false));
+
+    // Don't expose any protocol handlers by default and don't warn about those.
+    engineSettings->setPreference(QStringLiteral("network.protocol-handler.external-default"), false);
+    engineSettings->setPreference(QStringLiteral("network.protocol-handler.expose-all"), false);
+    engineSettings->setPreference(QStringLiteral("network.protocol-handler.warn-external-default"), false);
+
+    engine->rootContext()->setContextProperty("MozContext", webEngine);
```
chriadam: I guess we can't change the name of the context property at this point, yet.  Maybe add two: MozContext and WebEngine, then in "new" code (e.g. in this repo) use the WebEngine property?  Slowly migrate existing code (in Sailfish Browser) over to the WebEngine property at our leisure?

lib/webengine.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef SAILFISHOS_WEBENGINE_H
+#define SAILFISHOS_WEBENGINE_H
+
+#include <QString>
+#include <qmozcontext.h>
+
+namespace SailfishOS {
+
+class WebEngine : public QMozContext {
```
chriadam: Are we trying to get rid of QMozContext altogether at some stage?  Or not?  Just wondering if it makes sense to extend the public WebEngine class here, or instead provide some API and internally call through to the appropriate function of a QMozContext private data instance...

This is probably fine as-is, just thought I'd ask.

lib/webenginesettings.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <QtCore/QSize>
+#include <QtCore/QtMath>
+#include <QtGui/QGuiApplication>
+#include <QtGui/QScreen>
+
+Q_GLOBAL_STATIC(SailfishOS::WebEngineSettings, webEngineSettingsInstance)
+
+#define SAILFISH_WEBENGINE_DEFAULT_PIXEL_RATIO 1.5
+
+static bool testScreenDimensions(qreal pixelRatio) {
```
chriadam: prefer brace on separate line

lib/webenginesettings.cpp
```diff
@@ -0,0 +1,122 @@
...
+    QScreen *screen = QGuiApplication::primaryScreen();
+    qreal w = screen->size().width() / pixelRatio;
+    qreal h = screen->size().height() / pixelRatio;
+    return fmod(w, 1.0) == 0 && fmod(h, 1.0) == 0;
+}
+
+const QSettings &quickSettings()
+{
+    static const QSettings settings(QSettings::SystemScope, QStringLiteral("QtProject"), QStringLiteral("QtQuick2"));
```
chriadam: Why like this?  I would have expected UserScope with an applicationName derived from the application binary path (or read from /proc/pid etc)

chriadam:
A couple of minor comments / discussion points, but LGTM.

import/webengine/plugin.cpp
```diff
@@ -0,0 +1,49 @@
...
+    Q_OBJECT
+    Q_PLUGIN_METADATA(IID "Sailfish.WebEngine" )
+
+public:
+    void initializeEngine(QQmlEngine *, const char *)
+    {
+        // TODO : How to deal with custom default UA. We have also means to customize
+        // site specific UA overrides.
+        SailfishOS::WebEngine::initialize(QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+        SailfishOS::WebEngineSettings::initialize();
```
rainemak: Let's figure out once we have more use cases that how should we actually do the initialization. Maybe this is something that should not happen here at all rather we could have explicit QML API. On the other hand, this kind of approach would be quite convenient.

import/plugin.cpp
chriadam: If WebEngine is a QML singleton, the engine will own it (and delete it on engine dtor) which could cause issues if you try to use a Q_GLOBAL_STATIC instance.  But yes I agree with your plan, just implementation details we need to figure out.

chriadam:
LGTM

chriadam:
Other thing we need going forward (but shouldn't block this PR) is settings to enable or disable JavaScript and Image Loading

import/plugin.cpp
rainemak: We can already create instances of WebEngine/WebEngineSettings that will share private implementation. Problem that I'm currently seeing is commented in https://github.com/sailfishos/sailfish-components-webview/pull/7#pullrequestreview-697748 e.i. how to deal initialization?

Maybe need to just rid off initialization arguments and handle custom cases separately.

PR #8, opened 2016-09-21
[sailfish-components-webview] Provide WebViewFlickable to allow PullDown/PushUpMenu interaction. Contributes to JB#36097

chriadam:


chriadam:
My test application:

import QtQuick 2.1
import Sailfish.Silica 1.0
import Sailfish.WebView 1.0

ApplicationWindow {
    initialPage: WebViewPage {
        id: page
        anchors.fill: parent
        WebViewFlickable {
            anchors.fill: parent
            webView.url: "http://www.sailfishos.org"

            PullDownMenu {
                MenuItem { text: "itemOne" }
            }

            PushUpMenu {
                MenuItem { text: "itemTwo" }
            }
        }
    }
}

chriadam:
Updated PR and the example QML app.

import/webview/WebViewFlickable.qml
```diff
@@ -0,0 +1,85 @@
...
+            // As a hacky workaround, we disable flickable interaction for a brief
+            // period of time, giving the user a window of time within which to pan away,
+            // since touch events will not be delivered to the flickable but to the webview,
+            // however note that this requires the user to lift-finger and re-try.
+            //
+            // Note that the inverse does work properly without needing anything extra.
+            // That is, if you pan the webview to the top and then attempt to do a
+            // pull-down gesture (slowly) the pulley menu will eventually (after interactive
+            // gets reset back to "true" once the timer timeout expires) be activated,
+            // without requiring the user to lift their finger and try again.
```
chriadam: Raine, perhaps you could comment on this one?  To reproduce, follow these steps (with this PR installed):

1) wait for app to load fully, including rendering the web page in the web view

2) pan down somewhere in the webpage

3) pan back up to the top so that the WebView.atYBeginning will be true.  Note that at this point, due to the binding on flickable.interactive (which checks flickTimer.showPullDownMenu, which in turn relies on webView.atYBeginning), the flickable.interactive value will transition to "true" (that is, touch events are delivered to the Flickable and not the WebView).

4) If you then attempt to pan the webview downwards, this branch of the code should be hit (that is, a down-pan gesture is detected while flickable.interactive is true, and so the flickTimer will be started, making the Flickable non-interactive for one second).  Note, however, that this initial pan-down gesture DOES NOT pan the webview downwards (I assume it's because the initial touch events are delivered to the Flickable, and so even though it is then set interactive=false, the subsequent touch events aren't handled correctly by the WebView).

5) To get it to pan downwards, the user would have to lift their finger, and then start a second pan-downward gesture within one-second of the first one (else the timer would expire and they'd have to start again).  However, instead of doing that, do the following: (a) lift your finger, and then (b) attempt to pull-down the menu, very-very-slowly (i.e., make that pull-down gesture last longer than one second).  You will note that even though the "beginning" of the pull-down gesture will be delivered to the WebView (and hence, do nothing, since it's already panned to the atYBeginning position), after one second (when the timer expires and the Flickable is set interactive again) the latter touch events are delivered to the Flickable which causes the PullDownMenu to be activated.

6) My question is: why doesn't the inverse work (i.e., why doesn't the WebView end up handling the latter touch events from the initial down-pan gesture, after this codebranch is hit)?

rainemak: Maybe we could achieve the same by filtering events of flickable on webview (rawwebview).

Possible alternatives to deal this:

1) Replaying events that should move the content (the whole sequence of events (TouchBegin, TouchUpdate, and TouchEnd) on the event filter when we know that we can send those events through.

2) Synthesized touch begin could work from event filter.

3) Then there's also an API for moving web content (sendAsyncMessage("embedui:scrollTo", {"x": xValue, "y": yValue});

I think that 2nd could be actually rather good option for this.

chriadam: e.g. https://github.com/sailfishos/sailfish-components-webview/pull/10/files but I must be doing something wrong there, because the synthesised touch events don't cause the webview to pan in the required case.

import/webview/WebViewFlickable.qml
```diff
@@ -0,0 +1,85 @@
...
+            // I wish that the other way around "just worked" also...
+            flickTimer.start()
+        }
+    }
+
+    Timer {
+        id: flickTimer;
```
rainemak: Extra colon :)

PR #10, opened 2016-09-27
WIP: Touch-filter Flickable

chriadam:


chriadam:
This is similar to (based off) https://github.com/sailfishos/sailfish-components-webview/pull/8/files but taking into use Raine's idea of a touch event filter in the web view.

Unfortunately, I can't get the synthesised touch events to cause the webview to actually pan, and in general it seems to be much less smooth than the hackish Timer way from MR#8 but perhaps that's because something isn't working correctly due to a bug in my code?

PR #11, opened 2016-09-28
[sailfish-components-webview] Don't relayout content while VKB animates. Contributes to JB#36163

chriadam:
This commit adds a virtual keyboard observer to the WebView to ensure
that we don't relayout its content while the VKB animates.

Contributes to JB#36163

### DIFF
--- a/import/webview/WebView.qml
+++ b/import/webview/WebView.qml
@@ -182,7 +182,6 @@ RawWebView {
                 break
             }
             default: {
-                console.log("unhandled async message received: " + message)
                 break
             }
         }
@@ -260,22 +259,34 @@ RawWebView {
             return null
         }
 
+        function hasWebViewPage() {
+            return (webview.webViewPage != null && webview.webViewPage != undefined)
+        }
+
         function setActiveInPage() {
             if (webview.active) {
-                if (webview.webViewPage == null || webview.webViewPage == undefined) {
+                if (!hasWebViewPage()) {
                     webview.webViewPage = helper.findParentWithProperty(webview, '__sailfish_webviewpage')
                 }
-                if (webview.webViewPage != null && webview.webViewPage != undefined) {
+
+                if (hasWebViewPage()) {
                     webview.webViewPage.activeWebView = webview
                 }
+            }
+
+            if (!hasWebViewPage()) {
+                console.warn("WebView.qml it is mandatory to declare webViewPage property to get orientation change working correctly!")
             }
         }
     }
 
     SilicaPrivate.VirtualKeyboardObserver {
         id: virtualKeyboardObserver
+
+        readonly property QtObject appWindow: helper.findParentWithProperty(webview, "__silica_applicationwindow_instance")
+
         active: webview.enabled
-        transpose: __silica_applicationwindow_instance._transpose
+        transpose: appWindow ? appWindow._transpose : false
         onImSizeChanged: {
             if (imSize > 0 && opened) {
                 webview.virtualKeyboardMargin = virtualKeyboardObserver.imSize
--- a/import/webview/WebView.qml
+++ b/import/webview/WebView.qml
@@ -11,6 +11,7 @@
 
 import QtQuick 2.0
 import Sailfish.Silica 1.0
+import Sailfish.Silica.private 1.0 as SilicaPrivate
 import Sailfish.WebView 1.0
 import Sailfish.WebView.Popups 1.0
 import Sailfish.WebView.Pickers 1.0
@@ -267,6 +268,42 @@ RawWebView {
                 if (webview.webViewPage != null && webview.webViewPage != undefined) {
                     webview.webViewPage.activeWebView = webview
                 }
+            }
+        }
+    }
+
+    SilicaPrivate.VirtualKeyboardObserver {
+        id: virtualKeyboardObserver
+        active: webview.enabled
+        transpose: __silica_applicationwindow_instance._transpose
+        onImSizeChanged: {
+            if (imSize > 0 && opened) {
+                webview.virtualKeyboardMargin = virtualKeyboardObserver.imSize
+            }
+        }
+
+        orientation: {
+            if (webview.webViewPage != null) {
+                return webview.webViewPage.orientation
+            } else if (pageStack != undefined && pageStack != null) {
+                if (pageStack.currentPage !== undefined && pageStack.currentPage !== null) {
+                    return pageStack.currentPage.orientation
+                } else {
+                    return Orientation.Portrait
+                }
+            } else {
+                return Orientation.Portrait
+            }
+        }
+
+        onOpenedChanged: {
+            if (opened) {
+                webview.virtualKeyboardMargin = virtualKeyboardObserver.panelSize
+            }
+        }
+        onClosedChanged: {
+            if (closed) {
+                webview.virtualKeyboardMargin = 0
             }
         }
     }
--- a/import/webview/WebViewPage.qml
+++ b/import/webview/WebViewPage.qml
@@ -7,6 +7,45 @@ Page {
     property var activeWebView
 
     property int __sailfish_webviewpage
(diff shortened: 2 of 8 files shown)


import/webview/plugin.cpp
```diff
@@ -67,13 +67,40 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
...
+        m.setBottom(m.bottom() + delta);
+        setMargins(m);
+        emit virtualKeyboardMarginChanged();
+
+        QVariantMap map;
+        map.insert("imOpen", m_vkbMargin > 0);
+        map.insert("pixelRatio", SailfishOS::WebEngineSettings::instance()->pixelRatio());
+        map.insert("bottomMargin", m.bottom());
+        map.insert("screenWidth", width());
+        map.insert("screenHeight", height());
```
chriadam: This doesn't work.  I was hoping that something like "vkbMargin + height()" could be used to get the "original/non-vkb-open" height, but unfortunately it gives all sorts of bogus results when opening the VKB with different input method hints.

import/webview/WebView.qml
```diff
@@ -270,6 +271,35 @@ RawWebView {
         }
     }
 
+    SilicaPrivate.VirtualKeyboardObserver {
+        id: virtualKeyboardObserver
+        active: webview.enabled
+        orientation: {
+            if (webview.webViewPage != null) {
```
rainemak: Why not just like ```!webview.webViewPage```

import/webview/WebView.qml
```diff
@@ -270,6 +271,35 @@ RawWebView {
...
+                return webview.webViewPage.orientation
+            } else if (pageStack != undefined && pageStack != null) {
+                if (pageStack.currentPage !== undefined && pageStack.currentPage !== null) {
```
rainemak: ```return pageStack.currentPage ? pageStack.currentPage.orientation : Orientation.Portrait``` given that pageStack variable is defined.

import/webview/plugin.cpp
chriadam: To clarify: when opening a normal text input method VKB, height() here either returns 700 (with vkbMargin 580) or 700.56 (with vkbMargin 0) rather than 700 or 1280 (which is strange, but I guess it's an order-of-evaluation issue with QML).  But, when opening a numeric input method VKB, it can return different values such that height()+vkbMargin can be 1280, 1300, 1488, etc different values.  I don't understand why, but I guess it may not matter.

rainemak:
@chriadam amended your commit (hope you don't mind) and added one commit top.

Requires: https://git.merproject.org/mer-core/qtmozembed/merge_requests/25

@chriadam / @martinjones do you have chance to review?

OrientationFader is another component that can be shared between browser and webview.

import/webview/WebView.qml
```diff
@@ -259,15 +259,60 @@ RawWebView {
...
         }
 
+        function hasWebViewPage() {
+            return (webview.webViewPage == null || webview.webViewPage == undefined)
+        }
+
         function setActiveInPage() {
             if (webview.active) {
-                if (webview.webViewPage == null || webview.webViewPage == undefined) {
+                if (hasWebViewPage()) {
```
chriadam: this one should be: if !hasWebViewPage() I think

import/webview/WebView.qml
```diff
@@ -259,15 +259,60 @@ RawWebView {
...
                     webview.webViewPage = helper.findParentWithProperty(webview, '__sailfish_webviewpage')
                 }
                 if (webview.webViewPage != null && webview.webViewPage != undefined) {
```
chriadam: this line can be updated to: if hasWebViewPage() { ...

import/webview/WebView.qml
```diff
@@ -259,15 +259,60 @@ RawWebView {
             return null
         }
 
+        function hasWebViewPage() {
+            return (webview.webViewPage == null || webview.webViewPage == undefined)
```
chriadam: this check is the wrong way around, I think.  it should be: return (webview.webViewPage != null && webview.webViewPage != undefined)

import/webview/WebView.qml
```diff
@@ -259,15 +259,60 @@ RawWebView {
...
+            if (!hasWebViewPage()) {
+                console.warn("WebView.qml it is mandatory to declare webViewPage property to get orientation change working correctly!")
+            }
+        }
+    }
+
+    SilicaPrivate.VirtualKeyboardObserver {
+        id: virtualKeyboardObserver
+        active: webview.enabled
+        transpose: __silica_applicationwindow_instance._transpose
```
chriadam: is this an implicit resolution?  where is it defined?

import/webview/WebViewPage.qml
```diff
@@ -8,6 +8,32 @@ Page {
 
     property int __sailfish_webviewpage
 
+    orientationTransitions: orientationFader.orientationTransition
+
+    OrientationFader {
+        id: orientationFader
+
+        parent: __silica_applicationwindow_instance
```
chriadam: I really dislike implicit resolutions ;-)

import/webview/WebView.qml
rainemak: oops, sure :)

import/webview/WebView.qml
rainemak: Stupid copy paste error :). Will update.

import/webview/WebView.qml
```diff
@@ -259,15 +259,60 @@ RawWebView {
             return null
         }
 
+        function hasWebViewPage() {
+            return (webview.webViewPage != null && webview.webViewPage != undefined)
```
rainemak: Fixed this.

import/webview/WebViewPage.qml
```diff
@@ -7,6 +7,45 @@ Page {
     property var activeWebView
 
     property int __sailfish_webviewpage
+    default property alias _contentChildren: content.data
+
+    orientationTransitions: orientationFader.orientationTransition
+
+    Item {
+        id: content
```
rainemak: Added default content item. Now the fader is on top of this one.

chriadam:
Content layout is now done properly when the VKB opens, however try the following and flip the screen to landscape.  The rendering is squished horizontally rather than expanded, the touch points are translated strangely, and when the VKB opens the width is reduced again.

```import QtQuick 2.1
import Sailfish.Silica 1.0
import Sailfish.WebView 1.0

ApplicationWindow {
    allowedOrientations: Orientation.All
    initialPage: WebViewPage {
        id: page
        allowedOrientations: Orientation.All
        anchors.fill: parent
        WebViewFlickable {
            id: flickable
            anchors.fill: parent

            WebView {
                anchors.fill: parent
                url: "http://www.456bereastreet.com/lab/html5-input-types/"
                active: true
            } 

            PullDownMenu {
                MenuItem { text: "itemOne" }
            }

            PushUpMenu {
                MenuItem { text: "itemTwo" }
            }
        }
    }
}```

rainemak:
Fixes for date and time input types:

https://git.merproject.org/mer-core/embedlite-components/merge_requests/28 - fixes date/time input types

https://git.merproject.org/mer-core/qtmozembed/merge_requests/30 - doesn't really affect anything, just adds mapping

rainemak:
@chriadam / @martinjones how does this feel now? Now also "date" and "time" input types should work as expected.

chriadam:
I just dupped my device, rebuilt and redeployed sailfish-components-webview / popups / pickers packages, and retested with url: "http://www.456bereastreet.com/lab/html5-input-types/", and still had the same behaviour (squished in landscape mode).

Also, the date/time inputs seemed to just open the normal text-mode vkb - should a version --dup be enough to get all of the required updates, or do I need to do something else?

chriadam:
We can tackle the landscape support in a future MR. This LGTM.

PR #12, opened 2016-10-28
Make Flickable and WebView get along.

martinjones:
This removes WebViewPage as it is not necessary. It also includes the WebView in the WebViewFlickable, so only WebViewFlickable need be declared. It is available via the `webView` property.

It supports a footer only at the moment until setMargins() successfully sets top margins.

You'll notice that the sync of the footer and the web view is way out and this is due to the `scrollableOffset` updates being really infrequent. We need to get better update frequency in order to keep them more closely synced.

An example:

```
import QtQuick 2.1
import Sailfish.Silica 1.0
import Sailfish.WebView 1.0
import Qt5Mozilla 1.0

ApplicationWindow {
    initialPage: Page {
        id: page
        anchors.fill: parent

        WebViewFlickable {
            id: flickable
            anchors.fill: parent

                webView.url: "http://www.wikipedia.org"

                footer: Rectangle {
                        height: Theme.itemSizeLarge
                        color: "transparent"
                        Label {
                                anchors.centerIn: parent
                                text: "Footer"
                        }
                }

            PullDownMenu {
                MenuItem { text: "itemOne" }
            }

            PushUpMenu {
                MenuItem { text: "itemTwo" }
            }
        }
    }
}
```

### DIFF
--- a/import/webview/plugin.cpp
+++ b/import/webview/plugin.cpp
@@ -10,20 +10,17 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include "plugin.h"
+#include "rawwebview.h"
 #include "webengine.h"
 #include "webenginesettings.h"
 
-#include <QtCore/QTimer>
 #include <QtCore/QStandardPaths>
-#include <QtGui/QGuiApplication>
-#include <QtGui/QScreen>
 #include <QtQml/QQmlEngine>
 #include <QtQml/QQmlContext>
 
 #include <QQuickWindow>
 
 #define SAILFISHOS_WEBVIEW_MOZILLA_COMPONENTS_PATH QLatin1String("/usr/lib/mozembedlite/")
-#define CONTENT_ORIENTATION_CHANGED QLatin1String("embed:contentOrientationChanged")
 
 namespace SailfishOS {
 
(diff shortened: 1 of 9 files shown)


import/webview/plugin.cpp
```diff
@@ -67,13 +71,208 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
 
 RawWebView::RawWebView(QQuickItem *parent)
```
rainemak: Time to split RawWebView to own source file (.h/.cpp)?

rainemak:
@martinjones ah noticed that asyncscroll throttling is rather long. Following should help with scrollOffset updates:

Change webenginesettings.cpp line 63-64
```
engineSettings->setPreference(QStringLiteral("apz.asyncscroll.throttle"), QVariant::fromValue<int>(15));
engineSettings->setPreference(QStringLiteral("apz.asyncscroll.timeout"), QVariant::fromValue<int>(15));
```

rainemak:
Some comments:

- Web content is jumpy when slowly opening pushup menu -> feel a bit like there would be parallax effect :) => If I comment out scrollTo from onContentYChanged, then opening works as expected but closing pushup menu with the same gesture doesn't work.

- "ASSERT: "touchPointsCount > 0" in file qmozview_p.cpp, line 864" when selecting pulley menu item by tapping

- Don't understand when & how quick scroll indicator appears => worth disabling

- "ASSERT: "touchPointsCount > 0" in file qmozview_p.cpp, line 864" when tapping quick scroll indicator

martinjones:
> Change webenginesettings.cpp line 63-64

Yay! That is what I was looking for :)

> Web content is jumpy when slowly opening pushup menu -> feel a bit like there would be parallax effect :) => If I comment out scrollTo from onContentYChanged, then opening works as expected but closing pushup menu with the same gesture doesn't work.

I've tried to improve the sync, but there are still perf issues in this use case. If there's a way to improve the responsiveness of scrollTo() then that would help.

I got rid of the quick scroller, cleaned up a little and fixed the asserts.

rainemak:
Did you quick rebase and test. Now it looks and feels good.

"This removes WebViewPage as it is not necessary" => Added in another commit overlay on top of web content while applying orientation (OrientationFader). Needs to be amended to somewhere maybe directly to the WebView.

plugin.h / plugin.cpp has merge conflict:

1) virtualKeyboardMargin (property, signal, etc)
2) void contentOrientationChanged(Qt::ScreenOrientation orientation) signal
3) onAsyncMessage
4) and constructor

VKB margin and footer margin cannot conflict. So, maybe like vkbMargin + bottomMargin is what we should tell to the QtMozEmbed. Further, it might make sense that footer/header would follow item's visibility / opacity so that we'd remove footer/header margins when footer/header item is invisible. Visibility parts can be added later on if we feel that it is needed.

martinjones:
Rebased and cleaned up.

I've kept the WebViewPage. If we want to move the orientation change handing to somewhere else then that should be a separate change.

import/webview/rawwebview.h
```diff
@@ -0,0 +1,80 @@
...
+
+protected:
+    void handleTouchEvent(QTouchEvent *event);
+
+    void touchEvent(QTouchEvent *event);
+    bool childMouseEventFilter(QQuickItem *i, QEvent *e);
+
+signals:
+    void virtualKeyboardMarginChanged();
+    void contentOrientationChanged(Qt::ScreenOrientation orientation);    void flickableChanged();
```
rainemak: Move ```flickableChanged``` to own line

rainemak:
There's thin 1px line between footer and pullup menu when I open "http://www.wikipedia.org". However, that line is not visible when I open e.g.  "http://www.456bereastreet.com/lab/html5-input-types/". Looks like a rounding error somewhere that is not related to this. If you don't have idea from there it is coming from (not from this), let's file a task to this project. Quick entry -> Clone from can be used to copy it to internal bugzilla.

In addition to above there's one cosmetic issue. Fix that and remove WIPs before merging, LGTM.

martinjones:
Fixed the 1px line and fixed the joined lines in rawwebview.h.

PR #14, opened 2016-11-01
Jb36750

rainemak:
Cleanup PR and increases amount shareable code between QuickItem based WebView and browser. Defers component loading when applicable to avoid startup penalties. Didn a quick qml profiling:

For WebView.qml compilation was previously as follows:

- WebView.qml 920ms (biggest contributors below)

   - ContextMenu 230ms

   - PromptDialog 200ms

   - PickerCreator.qml 350ms

After:

- WebView.qml 160ms

   - PopupOpener 28ms

   - LocationSettings 24ms

   - VirtualKeyboardObserver 11ms

   - PickerOpener 8ms

Tested all popups and pickers with browser integration. Another PR appears shortly.

### DIFF
--- a/import/pickers/PickerOpener.qml
+++ b/import/pickers/PickerOpener.qml
@@ -1,0 +1,81 @@
+/****************************************************************************
+**
+** Copyright (C) 2016 Jolla Ltd.
+** Contact: Raine Makelainen <raine.makelainen@jolla.com>
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.2
+
+QtObject {
+    property var pageStack
+    property QtObject contentItem
+    readonly property var listeners: ["embed:filepicker", "embed:selectasync"]
+
+    // Defer compilation of picker components
+    readonly property string _multiSelectComponentUrl: Qt.resolvedUrl("MultiSelectDialog.qml")
+    readonly property string _singleSelectComponentUrl: Qt.resolvedUrl("SingleSelectPage.qml")
+    readonly property string _filePickerComponentUrl: Qt.resolvedUrl("PickerCreator.qml")
+    property Component _filePickerComponent
+
+    property var _messageHandler: Connections {
+        target: contentItem
+
+        onRecvAsyncMessage: {
+            if (!contentItem) {
+                console.warn("PickerOpener has no contentItem. Assign / Bind contentItem for each PickerOpener.")
+                return
+            }
+
+            if (!pageStack) {
+                console.log("PickerOpener has no pageStack. Add missing binding.")
+                return
+            }
+
+            var winid = data.winid
+            switch (message) {
+            case "embed:selectasync": {
+                pageStack.push(data.multiple ? _multiSelectComponentUrl : _singleSelectComponentUrl,
+                                               { "options": data.options, "webview": contentItem })
+                break
+            }
+            case "embed:filepicker": {
+                if (!_filePickerComponent) {
+                    _filePickerComponent = Qt.createComponent(_filePickerComponentUrl)
+                }
+
+                if (_filePickerComponent.status === Component.Ready) {
+                    _filePickerComponent.createObject(pageStack, {
+                                                          "pageStack": pageStack,
+                                                          "winid": winid,
+                                                          "webView": contentItem,
+                                                          "filter": data.filter,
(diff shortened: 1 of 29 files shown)


rainemak:
@chriadam review ?

import/webview/WebView.qml
```diff
@@ -31,19 +31,32 @@ RawWebView {
...
+    PopupOpener {
+        id: popupOpener
+
+        pageStack: pickerOpener.pageStack
+        parentItem: webview
+        contentItem: webview
     }
 
     onRecvAsyncMessage: {
+        if (pickerOpener.handlesMessage(message) || popupOpener.handlesMessage(message)) {
```
chriadam: One other possibility: instead of using a Connections object in the pickerOpener/popupOpener, just have a "function recvAsyncMessage(message, data, webview)" method, which this one calls if that one handlesMessage(message).  e.g.:

```
if (pickerOpener.handlesMessage(message)) {
    pickerOpener.recvAsyncMessage(message, data, webview);
} else if (popupOpener.handlesMessage(message) {
    popupOpener.recvAsyncMessage(message, data, webview);
} else {
    if (message == "embed:linkclicked") {
        // ...
    }
}
```

This has the advantage of also ensuring that the symbol resolutions in the pickerOpener / popupOpener are more explicit, since they would reference function parameters rather than the "hidden" signal parameters.

It's not important / it's just a matter of preference, but I thought I'd mention it.  It saves an object allocation too.

import/popups/LocationSettings.qml
```diff
@@ -0,0 +1,31 @@
...
+import com.jolla.settings.system 1.0
+import MeeGo.Connman 0.2
+
+QtObject {
+    id: root
+
+    property alias enabled: locationSettings.locationEnabled
+    property alias gpsPowered: gpsTechModel.powered
+
+    property LocationSettings l: LocationSettings {
```
chriadam: personal preference:
```
property LocationSettings locationSettings: LocationSettings {}
property TechnologyModel gpsTechModel: TechnologyModel { name: "gps" }
```

to have more descriptive property names and to avoid having unnecessary ids.

chriadam: Oh wait, then the property alias wouldn't work.  Never mind.

chriadam:
Some minor comments (just some different possibilities) but LGTM.

import/webview/WebView.qml
```diff
@@ -31,19 +31,31 @@ RawWebView {
...
-            "embed:login",
-            "Content:ContextMenu"]);
+                                        "embed:linkclicked",
+                                    ]);
+    }
+
+    PickerOpener {
+        id: pickerOpener
+
+        property QtObject pageStackOwner: helper.findParentWithProperty(webView, "pageStack")
```
rainemak: Not needed. Moved logging from PickerOpener Component.onCompleted to message handler. So, we do find pageStack but pagestack is unknown when PickerOpener's Component.onCompleted is called.

import/webview/WebView.qml
```diff
@@ -31,19 +31,32 @@ RawWebView {
...
+        contentItem: webview
     }
 
     onRecvAsyncMessage: {
+        if (pickerOpener.handlesMessage(message) || popupOpener.handlesMessage(message)) {
+            return
+        }
+
         // cache some symbol resolutions in var properties in the function closure
         var privData = helper
```
rainemak: Will add a commit on top that cleans up this and Timer parts. This I missed but originally didn't want to tamper ```helper``` now as this is growing fast.

import/webview/WebView.qml
rainemak: Did it like this, a bit inspired by QObject::event method:
```
if (pickerOpener.message(message, data) || popupOpener.message(message, data) {
    return;
}

```
Signal handler argument message is more like a topic and topic+data is the message. That's why message functions have prototype like ```message(topic, data)```

import/pickers/PickerOpener.qml
```diff
@@ -0,0 +1,83 @@
...
+                console.warn("PickerOpener failed to create PickerOpener: ", _filePickerComponent.errorString())
+            }
+            break
+        }
+        }
+        // If we end up here, message has been handled.
+        return true
+    }
+
+    function handlesMessage(topic) {
```
chriadam: instead of separate handlesMessage function, could just include this check directly in the message() function

chriadam:
One minor/optional comment, LGTM.

import/pickers/PickerOpener.qml
rainemak: Ah, didn't notice this one. I might clean it up later. Had like you suggested at some point but returned handlesMessage back.

PR #15, opened 2016-11-01
[sailfish-components-webview] Calculate "screen size" from current geometry. Contributes to JB#36163

rainemak:
Lanscape virtual keyboard handling was broken. This had nothing to
do with "date" and "time" input types that were mentioned in:

https://github.com/sailfishos/sailfish-components-webview/pull/11

### DIFF
--- a/import/webview/plugin.cpp
+++ b/import/webview/plugin.cpp
@@ -105,8 +105,12 @@ void RawWebView::setVirtualKeyboardMargin(qreal vkbMargin)
         map.insert("imOpen", m_vkbMargin > 0);
         map.insert("pixelRatio", SailfishOS::WebEngineSettings::instance()->pixelRatio());
         map.insert("bottomMargin", m_vkbMargin);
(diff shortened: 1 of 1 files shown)


rainemak:
@chriadam review ?

Version dup'ed my device, installed this, and simple WebView example. Now landscape vkb works better but these is still sometime a glitch when vkb is already and you move focus from one input element to another.

chriadam:
LGTM.  Is the glitch because the vkb for different input types might have different margin size?  Or something else?

rainemak:
Not 100% why the error occurs but my hunch is that error is lurking somewhere here. Wrote / adapted  that piece of code from Android FF like 2-3years ago:

https://git.merproject.org/mer-core/embedlite-components/blob/master/jsscripts/embedhelper.js#L387

rainemak:
As this fixes normal use case. Will merge this now. Filed https://github.com/sailfishos/sailfish-components-webview/issues/16 and added a copy to internal bugzilla.

PR #20, opened 2016-11-09
[sailfish-components-webview] Support header in WebViewFlickable. Contributes to JB#36749

chriadam:


rainemak:
LGTM

rainemak:
@adenexter I'd consider this as no go now. Do you agree?

adenexter:
Yes, it's sort of been incorporated without inline scrolling into merged changes.

PR #21, opened 2016-11-11
Jb36934

rainemak:
Add ShareTextPage and guard ViewplaceHolder.enabled with share method model ready ShareLinkPage 

@jpetrell review ?


jpetrell:
LGTM

PR #22, opened 2016-11-16
[sailfish-webengine-settings] Update text selection background color. Fixes JB#36973

rainemak:
Let's do not tamper QPalette just adjust perference.

@chriadam @jpetrell review. Tested some colors with Martin Schüle and we agreed go with "#878787" that's from roughly half way between white and black. This should cope rather well possible blending cases. Of course selection background can still blend with web content but that's totally different level problem.


jpetrell:
LGTM

rainemak:
Cleaned silica theme back to line where it was and fixed typo from commit message.

PR #23, opened 2016-11-16
[sailfish-components-webview] Share geolocation responses across tabs.  Fixes JB#36975

rainemak:
@jpetrell , @chriadam , @pvuorela review ?

### DIFF
--- a/import/popups/PopupOpener.qml
+++ b/import/popups/PopupOpener.qml
@@ -11,11 +11,11 @@
 
 import QtQuick 2.2
 import Sailfish.WebView.Popups 1.0 as Popups
+import "GeolocationHosts.js" as Geolocation
 
 Timer {
     id: root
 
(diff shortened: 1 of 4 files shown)


import/popups/PopupOpener.qml
```diff
@@ -11,11 +11,11 @@
 
 import QtQuick 2.2
 import Sailfish.WebView.Popups 1.0 as Popups
+import "GeolocationHosts.js" as Geolocation
 
 Timer {
     id: root
 
-    property var geolocationUrls: {{}}
```
rainemak: Regarding the issue this caused the undefined geolocationUrls errors. As I noticed that geolocation requests are not shared across tabs fixed that at the same go.

Should have been like:
```
property var geolocationUrls: ({})
```

chriadam:
LGTM

PR #24, opened 2016-11-29
[sailfish-components-webview] Cleanup invisible titles. Contributes to JB#19575

rainemak:
These are not used as DialogHeader titles. As we have search field
visible there as well I think that it is better to cleanup title from
MultiImagePickerDialog, MultiMusicPickerDialog, and MultiVideoPickerDialog
of sailfish-components-pickers rather than adding title here as well.

@jpetrell review ?


jpetrell:
LGTM

PR #27, opened 2016-12-23
[sailfish-components-webview] Always handle touch end. Fixes JB#37275

rainemak:
When a Silica Page containing WebView is peeked and gesture is ended,
we need to reset internal state. It does not matter if ended the gesture
results that the Page being popped from the PageStack or the Page
is returned back to the active state.


rainemak:
@martinjones @chriadam review ? Feel free to merge if you think that this is good way to fix the error.

rainemak:
@adenexter you have been looking this lately... Out of my head I'd say that this invalid now. Feel free to close.

adenexter:
I think the original issue was a consequence of the view doing event filtering, which is gone now and I don't think the problems been reproduced since that happened.

PR #29, opened 2017-03-07
[sailfish-components-webview] Don't use connman-jollagps plugin. Contributes to JB#37470

chriadam:


import/popups/PopupOpener.qml
```diff
@@ -99,28 +99,28 @@ Timer {
         }
         case "embed:permissions": {
             if (data.title === "geolocation"
-                    && Popups.LocationSettings.enabled
-                    && Popups.LocationSettings.gpsPowered) {
+                    && Popups.LocationSettings.locationEnabled
+                    && Popups.LocationSettings.gpsEnabled) {
                 switch (Geolocation.response(data.host)) {
                     case "accepted": {
```
chriadam: Should we also in this case trigger a QtPositioning.startUpdates() call here?  Or does something else handle actually getting the current location and returning it to the website (e.g., is there handling for that in embedlite?)

chriadam:
Hrm, actually, the connman-jollagps plugin allows Flight Mode to be set via connman (which will then unpower all radios including the GPS).  So perhaps this PR should be declined...

chriadam:
@rainemak regardless of whether we revert the second commit, I guess the first commit is still wanted?  I can split that out to a separate PR if you'd like, or we can wait to see what we do with the gps powered stuff first?

import/popups/PopupOpener.qml
```diff
@@ -99,28 +99,28 @@ Timer {
...
-                        contentItem.sendAsyncMessage("embedui:premissions",
+                        contentItem.sendAsyncMessage("embedui:permissions",
```
rainemak: Nice one, typo exists also over here: 

https://git.merproject.org/mer-core/embedlite-components/blob/master/jscomps/ContentPermissionPrompt.js#L114

https://git.merproject.org/mer-core/embedlite-components/blob/master/jscomps/ContentPermissionPrompt.js#L63

Could you prepare a separate PR about these typo fixes if this ends up being declined?

Regarding startUpdates, it's called over here.

https://git.merproject.org/mer-core/gecko-dev/blob/master/dom/system/qt/QTMLocationProvider.cpp#L66

chriadam: I created JB#38028 to track the typo fixes.

pvuorela:
LGTM

PR #34, opened 2017-08-29
[sailfish-components-webview] Use bigger tiles for FullHD and increase pixel ratio. Fixes JB#39557

rainemak:


rainemak:
@jpetrell @chriadam review ?

Applied pixel ratio for web content must result integer. Hence, those checks.

lib/webenginesettings.cpp
```diff
@@ -76,22 +76,25 @@ void SailfishOS::WebEngineSettings::initialize()
     // Round to nearest even rounding factor
     pixelRatio = qRound(pixelRatio / 0.5) * 0.5;
 
+    int screenWidth = QGuiApplication::primaryScreen()->size().width();
+
     // If we're on hdpi and calcaluted pixel ratio doesn't result integer dimensions, let's try to floor it.
```
pvuorela: typo :)

rainemak:
updated.

pvuorela:
As talked, still a chance that qRounded ratio gets wrong even if it was earlier right, but shouldn't break anything existing and can be adjusted again later.

LGTM.

PR #38, opened 2018-04-04
[components-webview] Use common sharing page. Contributes to JB#41068

pvuorela:
@rainemak

### DIFF
--- a/import/popups/ShareTextPage.qml
+++ b/import/popups/ShareTextPage.qml
@@ -13,40 +13,20 @@ import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Sailfish.TransferEngine 1.0
 
-Page {
+SharePage {
     id: root
 
     property string text
     property string type: "text/plain"
-    property string icon: "icon-launcher-browser"
 
-    ShareMethodList {
-        id: shareMethodList
-
-        anchors.fill: parent
-        header: PageHeader {
-            //: List header for sharing method list (generic)
-            //% "Share"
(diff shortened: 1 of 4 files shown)


import/popups/ShareTextPage.qml
```diff
@@ -13,40 +13,20 @@ import QtQuick 2.0
...
-        content: {
-            "name": text,
-            "data": text,
-            "type": type,
-            "icon": icon,
-            // also some non-standard fields for Twitter/Facebook status sharing:
-            "status" : text,
-            "linkTitle" : text
-        }
-        filter: type
```
rainemak: What happens/happened to this?

rainemak: Answering to myself mimeType is a sharemethodlist.filter alias.

import/popups/ShareTextPage.qml
```diff
@@ -13,40 +13,20 @@ import QtQuick 2.0
...
-        header: PageHeader {
-            //: List header for sharing method list (generic)
-            //% "Share"
-            title: qsTrId("sailfish_components_webview_popups-he-share")
-        }
-        content: {
-            "name": text,
-            "data": text,
-            "type": type,
-            "icon": icon,
```
rainemak: Don't recall but is this one of those deprecated fields?

pvuorela: Haven't spotted any usage and no one else seems to have idea why this has been there :)

PR #40, opened 2018-09-27
Expose PageStack pushMethod property from Popup-/PickerOpener

rainemak:
pushMethod is the name of the property. This allows utmost customizable page stack operation. As Browser has a custom background this makes it possible to map the background also from it.


rainemak:
Close in favour of https://github.com/sailfishos/sailfish-browser/pull/607

This would not have worked either.

PR #41, opened 2019-02-07
[sailfish-components-webview] Fixed pkgconfig version, contributes to jb#38551

mkenttala:
Maybe also tag as 1.0.0 so library file names stay compatible?


rainemak:
Yes, aim for keeping in sync keep with older. You could bump up already "Version: 0.0.15" from the spec.

rainemak:
LGTM :) +1

PR #44, opened 2019-05-17
[sailfish-components-webview] Correct license. Contributes to JB#45486

salmelamike:
For review

### DIFF
--- a/rpm/sailfish-components-webview.spec
+++ b/rpm/sailfish-components-webview.spec
@@ -1,5 +1,6 @@
 Name: sailfish-components-webview-qt5
-License: MPLv2
+License: MPLv2.0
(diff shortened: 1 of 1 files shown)


rpm/sailfish-components-webview.spec
```diff
@@ -1,5 +1,5 @@
 Name: sailfish-components-webview-qt5
-License: MPLv2
+License: MPLv2.0
```
rainemak: Url missing.

Please amend under the license:

```Url:        https://github.com/sailfishos/sailfish-components-webview```

So that column of the "h" of the https start from the same column as MPLv2.0

rainemak:
LGTM

PR #46, opened 2019-12-19
Add selection handles to WebView.

adenexter:
This takes the selection handles from sailfish-browser, adds them to a new TextSelection submodule, and integrates them into the WebView.

It appears to be doing the right things and a message is received from embedlite indicating text has been copied after selecting some text but there's nothing in the clipboard.

The change is top of the two outstanding merge requests which are varying degrees of necessary to work with the WebView right now.


lib/webenginesettings.cpp
```diff
@@ -110,6 +110,8 @@ void SailfishOS::WebEngineSettings::initialize()
     }
     engineSettings->setTileSize(QSize(tileSize, tileSize));
 
+    engineSettings->setPreference(QStringLiteral("browser.viewport.desktopWidth"), QVariant::fromValue(screenWidth));
```
adenexter: This produced some websites with crazy large fonts. I suspect the pixel ratio settings was compensating for what this is trying to fix and the combination blows up in some combinations, https://shop.jolla.com/ for example. A pixel ratio of 1 on a L500D looks about spot on, on an XA2 it's probably a bit squinty.

adenexter: I reduced the default pixel ratio to try and compensate but I'm clearly meddling with forces I don't understand as there's still an excess of scaling on some pages. Scaling the desktopWidth by the pixelRatio didn't seem to help either.

rainemak: Note that this these prefs affects also browser. Do you see difference there when comparing rendering result between browser and WebView?

rainemak: Further, you should not need to adjust desktop viewport value.

import/webview/WebViewFlickable.qml
```diff
@@ -1,53 +1,50 @@
+/****************************************************************************
+**
+** Copyright (c) 2016 Jolla Ltd.
```
rainemak: 2016 - 2020

import/webview/WebViewFlickable.qml
```diff
@@ -1,53 +1,50 @@
...
+** Copyright (c) 2020 Open Mobile Platform LLC.
+** Contact: Chris Adams <chris.adams@jollamobile.com>
```
rainemak: Please remove. ( sorry @chriadam  )

import/webview/WebViewPage.qml
```diff
@@ -1,65 +1,22 @@
+/****************************************************************************
+**
+** Copyright (c) 2016 Jolla Ltd.
+** Copyright (c) 2020 Open Mobile Platform LLC.
+** Contact: Chris Adams <chris.adams@jollamobile.com>
```
rainemak: Same

import/webview/WebViewPage.qml
```diff
@@ -1,65 +1,22 @@
+/****************************************************************************
+**
+** Copyright (c) 2016 Jolla Ltd.
```
rainemak: Same 2016 - 2020

import/textselection/TextSelectionController.qml
```diff
@@ -0,0 +1,237 @@
+/****************************************************************************
+**
+** Copyright (c) 2013 - 2019 Jolla Ltd.
+** Copyright (c) 2019 Open Mobile Platform LLC.
+** Contact: Dmitry Rozhkov <dmitry.rozhkov@jollamobile.com>
+** Contact: Raine Makelainen <raine.makelainen@jolla.com>
```
rainemak: Remove contacts

import/import.pro
```diff
@@ -1,5 +1,5 @@
 TEMPLATE=subdirs
-SUBDIRS+=webengine webview pickers popups
+SUBDIRS+=webengine webview pickers popups textselection
```
rainemak: Maybe this could a bit more generic subdir & import, controls maybe?

import/textselection/TextSelectionHandle.qml
```diff
@@ -0,0 +1,140 @@
+/****************************************************************************
+**
+** Copyright (C) 2013-2016 Jolla Ltd.
```
rainemak: Please change to lower-case (c) and "2013 - 2020 Jolla Ltd." and add:

** Copyright (c) 2020 Open Mobile Platform LLC.

And remove contacts

import/webview/WebView.qml
```diff
@@ -1,6 +1,7 @@
 /****************************************************************************
 **
-** Copyright (C) 2016 Jolla Ltd.
+** Copyright (c) 2016 Jolla Ltd.
+** Copyright (c) 2019 Open Mobile Platform LLC.
```
rainemak: 2020 for above and "2016 - 2020 Jolla Ltd."

import/webview/WebView.qml
```diff
@@ -15,12 +16,17 @@ import Sailfish.Silica.private 1.0 as SilicaPrivate
 import Sailfish.WebView 1.0
 import Sailfish.WebView.Popups 1.0
 import Sailfish.WebView.Pickers 1.0
+import Sailfish.WebView.TextSelection 1.0
```
rainemak: As said elsewhere maybe SailfishWebView.Controls import for the TextSelection and friends.

rpm/sailfish-components-webview.spec
```diff
@@ -64,6 +65,24 @@ Requires:  %{name} = %{version}
 %description pickers-ts-devel
 %{summary}.
 
+
+%package textselection
```
rainemak: controls could be merged also in package level (already now)

rpm/sailfish-components-webview.spec
```diff
@@ -85,15 +85,20 @@ Unit tests and functional tests of Sailfish OS WebView / WebEngine
...
 %{_datadir}/translations/sailfish_components_webview_qt5_eng_en.qm
+%{_datadir}/translations/sailfish_components_webview_controls_qt5_eng_en.qm
 %{_libdir}/qt5/qml/Sailfish/WebEngine/libsailfishwebengineplugin.so
 %{_libdir}/qt5/qml/Sailfish/WebEngine/qmldir
 %{_libdir}/qt5/qml/Sailfish/WebView/libsailfishwebviewplugin.so
 %{_libdir}/qt5/qml/Sailfish/WebView/qmldir
 %{_libdir}/qt5/qml/Sailfish/WebView/*.qml
+%{_libdir}/qt5/qml/Sailfish/WebView/Controls/libsailfishwebviewcontrolsplugin.so
+%{_libdir}/qt5/qml/Sailfish/WebView/Controls/qmldir
+%{_libdir}/qt5/qml/Sailfish/WebView/Controls/*.qml
```
rainemak: Could you align rpm package name as well.

import/import.pro
rainemak: Thanks

import/webview/WebView.qml
rainemak: Let's handle merging of Popups and Pickers to the Controls later.

rpm/sailfish-components-webview.spec
rainemak: also controls

import/controls/TextSelectionToolbar.qml
```diff
@@ -0,0 +1,100 @@
...
+        if (selectAllEnabled) {
+            buttons.push({ "icon": "image://theme/icon-m-select-all", "action": selectAll })
+        }
+
+        if (hasSearch) {
+            buttons.push({
+                "icon": "image://theme/icon-m-search",
+                "action": function () {
+                    search()
+                    clear()
```
rainemak: Regarding the review comment. It's better to first clear the search text and then trigger the action. Same applies for the other actions.

import/controls/TextSelectionToolbar.qml
```diff
@@ -0,0 +1,100 @@
...
+                }
+            })
+        }
+        buttons.push({
+            "icon": "image://theme/icon-m-input-remove",
```
rainemak: Is it intentional that this has changed?

adenexter: Yes.

> use just the cross icon on toolbar, without the background circle after text selection is invoked

Of course there are many cross icons so may not be the exact right one.

import/controls/TextSelectionToolbar.qml
adenexter: But then the text wouldn't be available for search.

rainemak: Browser side was previously like this:

                            controller.clearSelection()
                            webView.tabModel.newTab(controller.searchUri)
                            overlay.animator.showChrome(true)

Clearing selection did not clear yet the related search information.

adenexter: I've reversed the order so the clear() happens first. Irrespective of order though it was noted in another PR that clearing them immediately makes the icons change while a share page is being pushed so there's a little friction there.

PR #48, opened 2020-03-23
Move search engine configuration to WebEngineSettings.

adenexter:
This is primarily about ensuring all users of the web engine fulfill the
expectations of the EmbedLiteSearchEngine.js script which will otherwise
run up against a null object and throw a Type error.

### DIFF
--- a/lib/opensearchconfigs.cpp
+++ b/lib/opensearchconfigs.cpp
@@ -1,0 +1,79 @@
+/****************************************************************************
+**
+** Copyright (C) 2015
+** Contact: Siteshwar Vashisht <siteshwar@gmail.com>
+** Copyright (c) 2020 Open Mobile Platform LLC.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+#include <QDir>
+#include <QFile>
+#include <QXmlStreamReader>
+#include "opensearchconfigs.h"
+
+SailfishOS::OpenSearchConfigs *SailfishOS::OpenSearchConfigs::openSearchConfigs = 0;
+
+SailfishOS::OpenSearchConfigs::OpenSearchConfigs(QObject *parent)
+    : QObject(parent)
+{
+    m_openSearchPathList << QString(EMBEDLITE_CONTENT_PATH);
+    m_openSearchPathList << QDir::homePath() + USER_OPENSEARCH_PATH;
+}
+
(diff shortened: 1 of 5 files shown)


adenexter:
I'm not entirely sure we want to do this but I since I've gotten this far to trial whether the null _defaultEngine error is a crash trigger I might as well share it.

lib/opensearchconfigs.cpp
```diff
@@ -0,0 +1,78 @@
+/****************************************************************************
+**
+** Copyright (C) 2015
```
rainemak, line 3: Add Open mobile copyright as well.

lib/opensearchconfigs.h
```diff
@@ -0,0 +1,35 @@
+/****************************************************************************
+**
+** Copyright (C) 2015
```
rainemak, line 3: Same copyright line

lib/webenginesettings.cpp
```diff
@@ -21,6 +24,9 @@
 #include <QtGui/QScreen>
 #include <QtGui/QStyleHints>
 
+#include <MGConfItem>
+
```
rainemak, line 28: Extra empty line

lib/webenginesettings.h
```diff
@@ -25,6 +26,12 @@ class WebEngineSettings : public QMozEngineSettings {
 
     explicit WebEngineSettings(QObject *parent = 0);
     virtual ~WebEngineSettings();
+
+private:
+    QScopedPointer<WebEngineSettingsPrivate> d_ptr;
+
+    inline void setSearchEngine();
+    inline void handleObserve(const QString &message, const QVariant &data);
```
rainemak, line 34: Both of these could go to private class I think. Why inlining?

adenexter: Just an overly defensive habit, they're private members so the symbol visibility should also be private. Inline symbols aren't exported outside of the compilation unit. It's equivalent to a static non-member function or anonymous namespace in the .cpp file.

It's a little less over the top here though because WebEngineSettings should be an explicitly exported class so it also prevents the symbols being exported from the library. It seems no-ones configured explicit exports for this library though so everything is exported unless told otherwise.

adenexter: Putting them in the private class means the private class then needs a pointer back to the public class which is just more more complexity for no return.

rainemak: Thanks for clarification.

lib/webenginesettings.cpp
```diff
@@ -10,9 +10,12 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rainemak, line 10: Add copyright here as well.

rainemak:
LGTM

PR #49, opened 2020-04-02
[webview] Add a signal for when a page requests a link be opened in a new window. Contributes to JB#37998

adenexter:


import/webview/rawwebview.h
```diff
@@ -44,13 +47,16 @@ class RawWebView : public QuickMozView
     void virtualKeyboardMarginChanged();
     void contentOrientationChanged(Qt::ScreenOrientation orientation);
     void acceptTouchEventsChanged();
+    void openUrlInNewWindow(const QUrl &url);
```
rainemak, line 50: How are you planning to use this?

There are some cases where external links should be always opened to the browser application.

adenexter: The plan was to force target="_blank" on anchors, and then use to to invoke LinkHandler.handleLink(). It didn't work though because the target attribute didn't apply to implicit anchors.

So this doesn't solve my problem, but anchors that have target="_blank" are dead-ends in a web view at the moment so there might be some future value still.

import/webview/rawwebview.h
```diff
@@ -44,13 +47,16 @@ class RawWebView : public QuickMozView
...
 
 private:
     void onAsyncMessage(const QString &message, const QVariant &data);
 
+    std::shared_ptr<ViewCreator> m_viewCreator;
     qreal m_vkbMargin;
     QPointF m_startPos;
     bool m_acceptTouchEvents;
+
```
rainemak, line 59: Extra empty line.

import/webview/rawwebview.h
rainemak: Agree with possible future value.

This is something that needs to be handled regardless of the current issue that you're looking.

import/webview/rawwebview.cpp
```diff
@@ -14,31 +14,89 @@
...
+    , m_viewCreator(ViewCreator::instance())
     , m_vkbMargin(0.0)
     , m_acceptTouchEvents(true)
 {
+    m_viewCreator->views.push_back(this);
+
     addMessageListener(CONTENT_ORIENTATION_CHANGED);
 
     connect(this, &QuickMozView::recvAsyncMessage, this, &RawWebView::onAsyncMessage);
+
```
rainemak, line 94: Extra empty line

import/webview/rawwebview.cpp
```diff
@@ -14,31 +14,89 @@
...
 }
 
 RawWebView::~RawWebView()
 {
+    m_viewCreator->views.erase(std::find(m_viewCreator->views.begin(), m_viewCreator->views.end(), this));
```
rainemak, line 99: If we imagine that we have multiple webviews at the time, this clears a bit too much.

We can handle that separately when it becomes topical so I'd not worry about that now.

adenexter: I'm not sure why I used stl but that's the equivalent of QVector::remove(this).

import/webview/rawwebview.cpp
```diff
@@ -14,31 +14,89 @@
...
+{
+public:
+    ViewCreator();
+    ~ViewCreator();
+
+    quint32 createView(const QString &url, const quint32 &parentId) override;
+
+    static std::shared_ptr<ViewCreator> instance();
+
+    std::vector<RawWebView *> views;
```
rainemak, line 44: I think this should be set. having id and RawWebView pointer.

```std::set<quint32, RawWebView *>```

 Views without parent should be distinguishable from the views that are child views. A child view is a view that can be only closed from JavaScript as well whereas the parent view cannot.

Parent id is always zero (0). Views created from web content will always have > 0 id.

File: import/webview/rawwebview.cpp (hunk 4 above)
Comment (rainemak) [reply import/webview/rawwebview.cpp:99]: I like the stl.

Comment (monich) [reply import/webview/rawwebview.cpp:99]: With STL you can repeat `m_viewCreator->views` 3 times in a single statement, isn't that beautiful :)

Comment (rainemak) [reply import/webview/rawwebview.cpp:99]: Any updates?

File: import/webview/rawwebview.cpp (hunk 5 above)
Comment (adenexter) [reply import/webview/rawwebview.cpp:44]: The id would be the view's uniqueID though and that's not going to be available until the view is constructed which doesn't happen until sometime after the RawWebView constructor executes which means we'd have to notified when the view is constructed and only add the RawWebView to the global list of instances then.

File: import/webview/rawwebview.cpp (hunk 4 above)
Comment (adenexter) [reply import/webview/rawwebview.cpp:99]: On this snippet? No. The constructor appends a pointer to this view to a global list when it's constructed, this line removes that same pointer by finding it's position in the list and erasing the item at that position.

Comment (rainemak) [reply import/webview/rawwebview.cpp:99]: Misread this one previously. Sorry.

File: import/webview/rawwebview.cpp [hunk 6]
```diff
@@ -14,31 +14,89 @@
 #include "webengine.h"
 #include "webenginesettings.h"
 
+#include <qmozviewcreator.h>
```
rainemak, line 17: ** Copyright (c) 2016 - 2020 Jolla Ltd.
** Copyright (c) 2030 Open Mobile Platform LLC.

Note: Lower case c

rainemak:
LGTM

PR #50, opened 2020-04-20
[webengine] Don't create application caches by default. Contributes to JB#48786

adenexter:
The existence (or lack thereof) of a cache has a had an observed effect
on crash occurences. It's probably coincidental but it doesn't really
make sense and may even be detrimental to cache data to disk for most
web view uses which are mostly one-off transactions.

### DIFF
--- a/import/webengine/plugin.cpp
+++ b/import/webengine/plugin.cpp
@@ -34,7 +34,7 @@ public:
     {
         // TODO : How to deal with custom default UA. We have also means to customize
         // site specific UA overrides.
-        SailfishOS::WebEngine::initialize(QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+        SailfishOS::WebEngine::initialize(QString());
         SailfishOS::WebEngineSettings::initialize();
     }
 


adenexter:
Testing with afflicted devices has been promising.

import/webengine/plugin.cpp
```diff
@@ -34,7 +34,7 @@ class SailfishOSWebEnginePlugin : public QQmlExtensionPlugin
     {
         // TODO : How to deal with custom default UA. We have also means to customize
         // site specific UA overrides.
-        SailfishOS::WebEngine::initialize(QStandardPaths::writableLocation(QStandardPaths::CacheLocation));
+        SailfishOS::WebEngine::initialize(QString());
```
rainemak, line 37: To be noted that this shall not affect browser and it looks that it wouldn't. That said, maybe we should have some kind of global flag that could be used to disable profile.

adenexter: The browser is not affected because it initializes the web engine with it's own path (which is different to default given here) before anything loads the WebEngine import.

An application which wants to opt-in to a persistent cache can do the same.

rainemak: Yes, as said browser is not affected but this makes other use cases more cumbersome. At the same time, it's only matter of documenting the webengine from C++/QML. What the heck, let's have it like this for now and alter later if needed.

rainemak:
LGTM

PR #52, opened 2020-04-24
[webview] Anchor the webview busy indicator in a stable position. Contributes to JB#48894

adenexter:


pvuorela:
Not entirely sure how the cache location should be set, but in general LGTM.

adenexter:
The cache location is a straight revert to what it was before. Some amount of the javascript bundled in gecko throws exceptions when trying to access preferences if no cache path is set and it aborts execution of something selection handle related and probably other things so it seems as though it's necessary at this time.

import/controls/TextSelectionToolbar.qml
```diff
@@ -67,6 +82,25 @@ Row {
...
+    property int fontSize: {
+        var size = Theme.fontSizeTiny
+
+        if (!portrait && size > Theme.fontSizeTinyBase) {
+            var string = ""
+            for (var i = 0; i < buttons.length; ++i) {
+                string = string + (buttons[i].label || "")
+            }
+
+            var requiredWidth = metrics.advanceWidth(string) + (buttons.length * (Theme.itemSizeSmall + (2 * horizontalSpacing)))
```
pvuorela, line 94: Not sure I'm following the latter part. ItemSizeSmall maybe referring to icon, but the icon should be more of iconSizeMedium? Horizontal spacing is used between icon and label so it's only one time per button? 

String length concatenated vs individual parts summed might be slightly different because of kerning etc, but maybe doesn't matter that much.

adenexter: I thought it was iconSizeMedium too but that resulted in a number that was too small and the labels were all truncated which made me think I had to account for the item size instead because maybe the icon had a fixed size set, which was wrong but produced the right result. I'm still not sure what's going on there.

Twice per button is because there needs to be space between adjacent buttons too. But it's kind of irrelevant, I meant to set it to something smaller than paddingSmall but I forgot to initialize it and the spacing seemed adequate on testing perhaps because there's  a lot of padding in the graphic so I left it as 0. I could just take it out.

pvuorela:
The item size thing seems funny, but if there's anything it's erring on the safer side. Should be good enough for going forward: LGTM.

PR #55, opened 2020-05-23
[rpm] Set Version of pc file from rpm version

Thaodan:


rpm/sailfish-components-webview.spec
```diff
@@ -135,6 +136,8 @@ Unit tests and functional tests of Sailfish OS WebView / WebEngine
 
 %prep
 %setup -q -n %{name}-%{version}
+sed -i "s/^Version: .*/Version: "$(echo %{version} | sed 's/+.*//')"/" \
+    ./lib/pkgconfig/sailfishwebengine.pc
```
rainemak, line 140: Our integration system takes the tag from the git and uses that for the rpm package. That's the reason why we're "always" passing the VERSION from the spec down to the qmake (pc file generation). A bit depending on how are you building this could help you (SailfishOS-latest being the target name):

Use precisely version from the spec
> mb2 -t SailfishOS-latest -X build

Then without -X you get latest tag + extra identifier:
> mb2 -t Sailfish-latest build

rpm/sailfish-components-webview.spec
```diff
@@ -135,6 +136,8 @@ Unit tests and functional tests of Sailfish OS WebView / WebEngine
...
 
 %build
 %qmake5 VERSION=`echo %{version} | sed 's/+.*//'`
```
rainemak, line 143: So with above please just remove sed'ing from this line.

Thaodan: This means I can close this because the version is set elsewhere? No idea why I oversaw that.

rainemak: You can use this PR. Just drop sed from line 139-140 and this line. Then pc file matches to the lib regardless of the way how this was build.

rainemak: And edit a bit the commit title to better reflect the intention. Something like "Keep pc file version and library version in sync"

Thaodan: Like this?

rainemak: Yes. And now just drop your sed'ing from line 139-140 as well.

rpm/sailfish-components-webview.spec
rainemak: If you want to get this merged please drop this seding part.

This is enough:
```
%qmake5 VERSION=%{version}
```

Please rebase as well.

rainemak (changes requested):
No sed stuff...

Just:
```
%qmake5 VERSION=%{version}
```

okodron (approved):

rainemak:
@Thaodan could you please rebase this one.

PR #56, opened 2020-05-26
[webview] Remove "Cancel" from alert() dialogue. Contributes to JB#50018

llewelld:
According to W3Schools, a javascript alert() dialogue "displays an alert
box with a specified message and an OK button", in contrast with a
confirm() dialogue which also shows a Cancel button.

This change removes the "Cancel" option from the alert() dialogue.

### DIFF
--- a/import/popups/AlertDialog.qml
+++ b/import/popups/AlertDialog.qml
@@ -18,6 +18,8 @@ UserPrompt {
     //: Text on the Accept dialog button that accepts browser alert messages
     //% "Ok"
     acceptText: qsTrId("sailfish_components_webview_popups-he-accept_alert")
+    // Cancel button left blank on the alert() dialog
+    cancelText: ""
 
     PromptLabel {
         id: label
--- a/import/popups/UserPrompt.qml
+++ b/import/popups/UserPrompt.qml
@@ -15,6 +15,7 @@ import Sailfish.Silica 1.0
 Dialog {
     id: dialog
     property alias acceptText: header.acceptText
+    property alias cancelText: header.cancelText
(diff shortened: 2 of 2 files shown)


import/popups/AlertDialog.qml
```diff
@@ -19,6 +19,8 @@ UserPrompt {
     //% "Ok"
     acceptText: qsTrId("sailfish_components_webview_popups-he-accept_alert")
 
+    backNavigation: false
```
rainemak, line 22: Nope... If you like to "fix" this like this... I'd suggest just rename "Cancel" to "Ok". This does not affect to the reply messages that we're sending back to the engine.

llewelld: Could you elaborate on your reasons please? Do you mean have two buttons with "OK" on them?

rainemak: Sure... It's just dismissing it from both edges. As we're using dialogs (pages) for this would be good not to break navigation patterns (having two OK "buttons" is not bad).

That said, using dialogs for this is just plain wrong as it should be bound to the context where is was triggered not to some platform navigation logic. Further, a fullscreen dialog pulls user's focus away from the context where the alert was triggered this way causing unnecessary context switches (confusion).

rainemak: Any updates?

llewelld: I've updated the PR to allow left-swiping again. Martin agreed this should still be supported. He suggested to just remove the text from the "Cancel" button, which this PR now reflects.

rainemak:
LGTM

PR #58, opened 2020-06-05
[sailfish-webview] Align authentication dialog properties. Fixes JB#50121

rainemak:


rainemak:
Requires: https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/59

rainemak:
Will bump up version and requirements after this is approved.

import/popups/AuthDialog.qml
```diff
@@ -60,23 +68,36 @@ UserPrompt {
...
+            width: parent.width - Theme.paddingLarge * 2
+            visible: privateBrowsing
+            wrapMode: Text.Wrap
+            color: Theme.highlightColor
 
-            //: Explain to the user that checking the switch will prevent entered credentials from saving
-            //% "The credentials you enter won't be stored for later use"
-            description: qsTrId("sailfish_components_webview_popups-la-dont_save_description")
+            //: Description label for user when private mode active (entered credentials are not saved)
+            //% "The entered credentials won't never be stored in private browsing mode"
```
abranson, line 99: It should be "won't ever", but I suggest a drier:
"Credentials cannot be stored in private browsing mode"

rainemak: Technically we probably could  store then. Hence, I'd put more emphasis to the behaviour. Should be it like:
"Credential are not stored in private browsing mode"

abranson: The 'cannot' is more of a 'this program does not make it possible' sort of thing. 'Are not' would suggest that the storage has no effect. Could go with an unambigious 

'Credential storage is disabled in private browsing mode'

rainemak: Yeah that sounds better. Maybe "not disabled"... but "not available"? How does that sound like to you?

abranson: Yes, that's better. 'Disabled' might make people think they can re-enable it in some setting somewhere.

rainemak: Now amended. Thanks @abranson for this ping-pong and formatting this one.

PR #61, opened 2020-07-21
[sailfish-webview] Add permission manager. Contributes to JB#50492

atatarov:


atatarov:
Example:
`
PermissionModel {
     Component.onCompleted: PermissionManager.requestAllPermissions()
     Connection {
          target: PermissionManager
          onReplyAll: init(PermissionManager.allPermissions)
     }
}
`

llewelld:
A few of the files need copyright headers still.

llewelld:
The code here all looks decent, but I'm unsure about the overall architecture. It's hard to really understand whether this will cover what's needed without also seeing at least the start of the UI implementation to be built on top of it. As we discussed elsewhere, I'm especially uncertain about whether this minimal model, containing just the URLs, will be enough.

For example it might be better to include the permission settings for each URL in the model as well. This may be useful for the UI, but also makes the model more generally useful across the UI too.

What's the prospect of seeing a minimal UI implementation in the near future? I think it would be ideal if we could test this out properly before merging.

atatarov:
PermissionModel is final and needs a review. It is necessary to add a filter for the model, perhaps QSortFilterProxyModel and think about how to filter it
- filter by type (e.g. "geolocation")
- filter by unique uri

import/permissions/permissionsplugin.cpp
```diff
@@ -0,0 +1,45 @@
+/****************************************************************************
```
rainemak, line 1: No need to introduce new import for this one. At least, this is getting too detailed. We intentionally used broad import for text selection called "Controls". I think that this can be seen a control as well.

import/webview/WebView.qml
```diff
@@ -97,7 +98,9 @@ RawWebView {
     }
 
     onRecvAsyncMessage: {
-        if (pickerOpener.message(message, data) || popupOpener.message(message, data)) {
+        if (pickerOpener.message(message, data)
+                || popupOpener.message(message, data)
+                || PermissionManager.message(message, data)) {
```
rainemak, line 103: I guess we need to opt in this handling here due to in app usage and we don't necessarily want to handle permissions. See other comments as well.

rainemak: Further this looks none view specific on the https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/63/diffs#4eb83ed266b2098cf49b438ad1f855e80bdafa27_0_36

WebEngine's notifyObservers and addObserver(s) should be enough here.

import/permissions/permissionsplugin.h
```diff
@@ -0,0 +1,60 @@
...
+    void initializeEngine(QQmlEngine *engine, const char *uri);
+};
+
+// using custom translator so it gets properly removed from qApp when engine is deleted
+class AppTranslator: public QTranslator
+{
+    Q_OBJECT
+
+public:
+    AppTranslator(QObject *parent)
```
llewelld, line 42: Is the translation code necessary? There don't seem to be any translations here (or potential need in the future).

import/webview/WebView.qml
atatarov: Looked at notifyObservers and did not find how to use it to catch the data signal when requesting the list of permissions, could you give more information please

rainemak: There's  "void recvObserve(const QString message, const QVariant data);" signal in the WebEngine. You can connect to that in the model itself. In addition, you should call in the model WebEngine::addObserver("mytopic"); to register listening messages.

import/controls/permissionmodel.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+void PermissionModel::remove(int index)
+{
+    beginRemoveRows(QModelIndex(), index, index);
+    m_permissionList.removeAt(index);
+    endRemoveRows();
+}
+
+void PermissionModel::setCapability(QModelIndex index, QString capabacity)
```
llewelld, line 87: Just a minor change, but this looks like a typo (`capabacity` vs. `capability`).

atatarov: i'm sorry, it's my typo, i would like to use "capability"

import/controls/permissionmanager.cpp
```diff
@@ -0,0 +1,52 @@
+#include "permissionmanager.h"
```
llewelld, line 1: I've just noticed that this file, and import/controls/permissionmanager.h still need copyright headers.

llewelld (changes requested):

llewelld (changes requested):
A couple of files still need copyright headers.

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <QObject>
+
+class PermissionManager : public QObject
+{
+    Q_OBJECT
+public:
+    PermissionManager(QObject *parent = nullptr);
+
+    // permission cases: 'unknown', 'allow', 'deny', 'prompt'
```
rainemak, line 12: These should be declared as enums.

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,30 @@
...
+    Q_INVOKABLE void add(const QString &uri, const QString &type, const QString &permission);
+    Q_INVOKABLE void remove(const QString &uri, const QString &type);
+    Q_INVOKABLE void removeAll();
+    Q_INVOKABLE void requestUriPermissions(const QString &uri);
+
+    static void sendRequest(const QString &message
+                       , const QString &uri = ""
+                       , const QString &type = ""
+                       , const QString &permission = "");
```
rainemak, line 21:
```
    static void sendRequest(const QString &message,
                       const QString &uri = QString(),
                       const QString &type = QString(),
                       const QString &permission = QString());
```

import/controls/permissionmodel.h
```diff
@@ -0,0 +1,57 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef PERMISSIONMODEL_H
+#define PERMISSIONMODEL_H
+
+#include <QObject>
```
okodron, line 14: Not necessary here

import/controls/permissionmodel.h
```diff
@@ -0,0 +1,57 @@
...
+        Capability
+    };
+
+    PermissionModel(QObject *parent = nullptr);
+
+    Q_INVOKABLE void remove(int index);
+    void setCapability(QModelIndex index, QString capability);
+
+    QVariant data(const QModelIndex &index, int role = Qt::DisplayRole) const override;
+    int rowCount(const QModelIndex &parent = QModelIndex()) const override;
```
okodron, line 44: What about `columnCount()`?

import/controls/permissionmodel.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef PERMISSIONMODEL_H
+#define PERMISSIONMODEL_H
+
+#include <QObject>
+#include <QAbstractListModel>
+
+struct Permission
+{
+    Permission(QString uri, QString type, QString capability)
```
okodron, line 19: I guess this struct should depict the original [`nsIPermission`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPermission). If it's so then let's use the same names: `m_host` instead of `m_uri`.
Also it would be nice to leave a small comment what's type and what's capability.

import/controls/permissionmanager.h
okodron: Indeed. Better to store this value as an integer instead of a string constant. The same is about the `embedlite-components` part - I think it's possible to send the integer value instead of the string.

okodron (changes requested):

import/controls/permissionmodel.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "webengine.h"
+#include "permissionmanager.h"
+
+PermissionModel::PermissionModel(QObject *parent) : QAbstractListModel(parent)
+{
+    SailfishOS::WebEngine *webEngine = SailfishOS::WebEngine::instance();
+
+    connect(webEngine, &SailfishOS::WebEngine::recvObserve, this, &PermissionModel::handleRecvObserve);
+
+    webEngine->addObserver(QStringLiteral("embed:perms:all"));
```
okodron, line 21: I think it could be defined once as a constant, also used at the line 60.

import/controls/permissionmanager.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "webengine.h"
+
+PermissionManager::PermissionManager(QObject *parent)
+    : QObject(parent)
+{
+    SailfishOS::WebEngine *webEngine = SailfishOS::WebEngine::instance();
+
+    connect(webEngine, &SailfishOS::WebEngine::recvObserve, this, &PermissionManager::handleRecvObserve);
+
+    webEngine->addObserver(QStringLiteral("embed:perms:all-for-uri"));
```
okodron, line 11: Can be used as a constant.

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,30 @@
...
+
+signals:
+    void finished(const QVariant &data);
```
okodron, line 24: You should leave a comment that it's the result of `requestUriPermissions()` call.

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,30 @@
...
+class PermissionManager : public QObject
+{
+    Q_OBJECT
+public:
+    PermissionManager(QObject *parent = nullptr);
+
+    // permission cases: 'unknown', 'allow', 'deny', 'prompt'
+    Q_INVOKABLE void add(const QString &uri, const QString &type, const QString &permission);
+    Q_INVOKABLE void remove(const QString &uri, const QString &type);
+    Q_INVOKABLE void removeAll();
```
okodron, line 15: I think these methods are static as well.

import/controls/permissionmodel.cpp
```diff
@@ -0,0 +1,92 @@
...
+                                           varMap.value("type").toString(),
+                                           varMap.value("capability").toString()));
+    }
+    endResetModel();
+}
+
+void PermissionModel::remove(int index)
+{
+    beginRemoveRows(QModelIndex(), index, index);
+    m_permissionList.removeAt(index);
```
okodron, line 83: Maybe you should call `PermissionManager::remove()` otherwise the model just hides an entry.

okodron (changes requested):

import/controls/permissionmodel.h
```diff
@@ -0,0 +1,65 @@
...
+#ifndef PERMISSIONMODEL_H
+#define PERMISSIONMODEL_H
+
+#include <QAbstractListModel>
+#include "permissionmanager.h"
+
+struct Permission
+{
+    Permission(QString host, QString type, PermissionManager::Capability capability)
+        : m_host(host), m_type(type), m_capability(capability) {}
```
rainemak, line 20:
```
    Permission(QString host, QString type, PermissionManager::Capability capability)
        : m_host(host)
        , m_type(type)
        , m_capability(capability)
    {
    }
```
In contrast to the splitting parameters of a method over multiple lines. This way it is easier to see when ever somebody adds a new variable to the constructor list.

import/controls/permissionfilterproxymodel.cpp
```diff
@@ -0,0 +1,68 @@
...
+{
+
+}
+
+void PermissionFilterProxyModel::remove(int currentIndex)
+{
+    QModelIndex proxyIndex = index(currentIndex, 0);
+    QModelIndex sourceIndex = mapToSource(proxyIndex);
+
+    PermissionModel *permissionModel = dynamic_cast<PermissionModel *>(sourceModel());
```
rainemak, line 26: Use qobject_cast<PermissionModel *>(sourceModel()); instead.

import/controls/permissionfilterproxymodel.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+    permissionModel->remove(sourceIndex.row());
+}
+
+void PermissionFilterProxyModel::setCapability(int currentIndex, int capability)
+{
+    QModelIndex proxyIndex = index(currentIndex, 0);
+    QModelIndex sourceIndex = mapToSource(proxyIndex);
+
+    PermissionModel *permissionModel = dynamic_cast<PermissionModel *>(sourceModel());
```
rainemak, line 38:
```suggestion
    PermissionModel *permissionModel = qobject_cast<PermissionModel *>(sourceModel());
```
RTTI should be avoided.

import/controls/permissionfilterproxymodel.cpp
rainemak, line 26:
```suggestion
    PermissionModel *permissionModel = qobject_cast<PermissionModel *>(sourceModel());
```

rainemak (changes requested):
@atatarov could check the comments. Let's get this one integrated and fix whatever needed on follow up PRs.

import/controls/permissionfilterproxymodel.cpp
```diff
@@ -0,0 +1,68 @@
+/****************************************************************************
+**
+** Copyright (C) 2020 Open Mobile Platform LLC
```
rainemak, line 3:
```suggestion
** Copyright (c) 2020 Open Mobile Platform LLC.
```
Lower case c and ending with .

rainemak (approved):
Thanks for the final push! LGTM

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,53 @@
...
+class Permission;
+
+class PermissionManager : public QObject
+{
+    Q_OBJECT
+public:
+    PermissionManager(QObject *parent = nullptr);
+
+    enum Capability {
+        Unknown,
```
okodron, line 27: I would recommend to set enum values to the same as in https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPermissionManager and leave a comment to keep this manual synchronization in mind.

rainemak: Very good point. Add both mozilla doc ref and https://git.sailfishos.org/mer-core/gecko-dev/blob/master/netwerk/base/nsIPermissionManager.idl ref

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,53 @@
...
+        Allow,
+        Deny,
+        Ask
```
rainemak, line 30: And due to above this should be "Prompt" I'd say.

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,53 @@
...
+#include <QScopedPointer>
+
+class Permission;
+
+class PermissionManager : public QObject
+{
+    Q_OBJECT
+public:
+    PermissionManager(QObject *parent = nullptr);
+
```
rainemak, line 25:
```suggestion
    // See https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPermissionManager
    // And https://git.sailfishos.org/mer-core/gecko-dev/blob/master/netwerk/base/nsIPermissionManager.idl 
```

import/controls/permissionmanager.h
rainemak, line 27:
```suggestion
        Unknown = 0,
```

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,53 @@
...
+    enum Capability {
+        Unknown,
+        Allow,
```
rainemak, line 28:
```suggestion
        Allow = 1,
```

import/controls/permissionmanager.h
```diff
@@ -0,0 +1,53 @@
...
+        Deny,
```
rainemak, line 29:
```suggestion
        Deny = 2,
```

import/controls/permissionmanager.h
rainemak, line 30:
```suggestion
        Prompt = 3
```

llewelld (approved):

PR #62, opened 2020-08-07
Increase touch responsiveness

llewelld:
Applies the following default settings:

```
apz.content_response_timeout: 600
layout.css.touch_action.enabled: false
```

This has the effect of increasing the likelihood of touch events being
passed to the correct content item, while also removing the lag when
dragging the page with a finger.


llewelld:
These settings will be picked up directly from the [changes to embedding.js in gecko-dev](url) (already merged). The only reason they weren't being picked up originally is that I'd failed to add semicolons to the end of the preference lines.

So this PR is redundant and I'm closing it.

PR #66, opened 2020-09-01
[sailfish-components-webview] Make DownloadHelper::createUniqueFileUrl more safe. Contributes to JB#50923

paveltkv:
- Added replacement of illegal and whitespace characters with dashes and underscores;
- File name length is limited to 255 symbols;
- Added tests.

### DIFF
--- a/lib/downloadhelper.cpp
+++ b/lib/downloadhelper.cpp
@@ -1,6 +1,7 @@
 /****************************************************************************
 **
 ** Copyright (C) 2016 Jolla Ltd.
+** Copyright (c) 2020 Open Mobile Platform LLC.
 ** Contact: Raine Makelainen <raine.makelaine@jolla.com>
 ** Contact: Chris Adams <chris.adams@jolla.com>
 **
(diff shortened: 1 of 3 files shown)


lib/downloadhelper.cpp
```diff
@@ -14,29 +15,46 @@
...
-    QString result(path + "/" + fileName);
-    int collisionCount(1);
+    // Replace whitespace characters with underscores
+    fileName.replace(QRegExp("[\\s_-]+"), "_");
 
-    while (QFileInfo::exists(result)) {
-        collisionCount++;
-        result = newFile.arg(path).arg(baseName).arg(collisionCount).arg(suffix.isEmpty() ? "" : ".").arg(suffix);
+    // Determine of extension position
+    int dotPosition = fileName.length();
```
rainemak, line 39: lastIndexOf('.')

rainemak: Actually even better is to use QFileInfo and completeSuffix so that combined extensions are covered as well .tag.gz like cases.

lib/downloadhelper.cpp
```diff
@@ -14,29 +15,46 @@
...
+    for (int i = 0, j = fileName.length() - 1; i < fileName.length() && i < FILEEXTENSION_MAX_LENGTH; i++, j--) {
+        if (fileName[j] == '.') {
+            dotPosition = j;
+        }
     }
 
+    const QString baseName = fileName.left(dotPosition);
+    const QString extension = fileName.mid(dotPosition);
```
rainemak, line 47: QFileInfo baseName and completeSuffix

lib/downloadhelper.cpp
```diff
@@ -14,29 +15,49 @@
...
+
+    QString result = "";
```
pvuorela, line 52: Unnecessary assignment.

lib/downloadhelper.cpp
paveltkv: If pass a file name of, for example, 300 characters, with a leading dot, completeSuffix will return it in its entirety. This is incorrect, since there are no such long extensions. And this does not allow shortening the filename to 255 symbols, since only basename can be shortened, which is empty in this case. Limiting the length of the extension to some reasonable value solves this problem.

rainemak: Leading dot and your case is not resulting right extension either -- it's just something that is fabricated.

Not true either with one leading dot suffix is empty. When there are n leading dots then completeSuffix is n-1 dots. This can be consider valid as well. But if you think that leading dots are not good... then something like this would do it (wrote without compiling):

```
while (fileName.endsWith('.') && fileName.length() > 0) {
    fileName.chop(1);
}

if (fileName.isEmpty() {
    fileName = QStringLiteral("unknown_file"); // or something like
}
const QFileInfo fileInfo(fileName);
QString baseName = fileInfo.baseName();
const QString suffix = fileInfo.completeSuffix();

int maxLen = 255 - suffix.length();
baseName = baseName.left(maxLen);
```

adenexter: Look into QMimeDatabase. It has a suffixForFileName() function which should return the correct suffix for known extensions, and if the mime type is known but there's no extension in the file name it should be able to provide one.

paveltkv:
Fixed in PRs: https://github.com/sailfishos/sailfish-components-webview/pull/81 https://github.com/sailfishos/sailfish-components-webview/pull/83 https://github.com/sailfishos/sailfish-components-webview/pull/88 https://github.com/sailfishos/sailfish-components-webview/pull/89

PR #67, opened 2020-09-03
[webview] Load js libraries from the correct location. JB#50653

krnlyng:


okodron (approved):

abranson (approved):
Great catch!

rainemak (approved):
LGTM as well

rainemak:
Btw there similar path in the sailfish-browser. Maybe you'd like to fix that as well.

PR #70, opened 2020-09-14
[webview] Limit popup message length. JB#51217

shanechko:


import/popups/PopupOpener.qml
```diff
@@ -62,6 +62,11 @@ Timer {
             return false
         }
 
+        if (data.text.length > 300) {
```
rainemak, line 65: I'm ok with this one. Let's just increase the length a bit. Maybe something like 1000? Any opinions?

import/popups/PopupOpener.qml
```diff
@@ -62,6 +62,11 @@ Timer {
...
+            console.log("Message is too big to show and will be truncated.")
+            data.text  = data.text.substring(0,300) + " ..."
```
rainemak, line 67: space missing after comma (0, 300)

import/popups/PopupOpener.qml
rainemak: + if (data.text &&...

shanechko: done

import/popups/PopupOpener.qml
shanechko: ok

rainemak (approved):
LGTM

rainemak:
Sorry. Didn't notice that this went to wrong branch. Will cherry-pick. The "next" branch is always the next coming version.

PR #74, opened 2020-10-14
[webview] Backport notification changes to 3.4.0. Contributes to JB#51257

adenexter:


rainemak:
Will push the whole master to the upgrade-3.4.0 soonish.

adenexter:
Okay... This change has a dependency to Silica which needs to go in first and the version dependencies in the spec file have to be lowered to match the upgrade-3.4.0 branch there.

rainemak:
Note also that there was also some background change fix that needs to go the to Silica.

pvuorela:
Suppose not needed if the whole branch is pushed, but for completeness: LGTM :)

PR #79, opened 2020-11-06
[sailfish-components-webview] Always display geolocation prompt dialog when geolocation permission is requested. Contributes to JB#51932

paveltkv:
…


rainemak (approved):
My concern is that location request dialog might be too annoying if the location is disabled. That said, let's give this a spin / chance and see how does this work in daily use.

PR #80, opened 2020-11-18
[sailfish-webview] Add for single host model case. Add Expiration enum. Contributes to JB#50927

atatarov:
Add m_host property and update the model according to its value.
If m_host is empty request all permissions. Add the expiration role to the model.


import/controls/permissionfilterproxymodel.h
```diff
@@ -28,11 +29,16 @@ class PermissionFilterProxyModel : public QSortFilterProxyModel
...
+    bool onlyPermanent() const;
+    void setOnlyPermanent(bool onlyPermanent);
+
 signals:
     void permissionTypeChanged(QString permissionType);
+    void onlyPermanentChanged(bool onlyPermanent);
 
 private:
     QString m_permissionType;
+    bool m_onlyPermanent;
```
okodron, line 41: Missing initialization in `PermissionFilterProxyModel` constructor.

llewelld:
This looks like a nice change. I'll test it against sailfish-browser PRs [725](https://github.com/sailfishos/sailfish-browser/pull/725) and [728](https://github.com/sailfishos/sailfish-browser/pull/728).

One small point is that our style guide (which I'm not sure you have direct access to?) requires braces even around single-line conditionals:

> **2-6. Use braces in one-line conditionals**
> write
> ```
> if (a) {
>     doSomething()
> }
> ```
> instead of
> ```
> if (a)
>     doSomething()
> ```
There are a number of places in the code where the latter is used.

atatarov:
> This looks like a nice change. I'll test it against sailfish-browser PRs [725](https://github.com/sailfishos/sailfish-browser/pull/725) and [728](https://github.com/sailfishos/sailfish-browser/pull/728).
> 
> One small point is that our style guide (which I'm not sure you have direct access to?) requires braces even around single-line conditionals:
> 
> > **2-6. Use braces in one-line conditionals**
> > write
> > ```
> > if (a) {
> >     doSomething()
> > }
> > ```
> > 
> > 
> > instead of
> > ```
> > if (a)
> >     doSomething()
> > ```
> 
> There are a number of places in the code where the latter is used.

please, cooment it https://github.com/sailfishos/sailfish-browser/pull/747
It use 3rd case for this model

import/controls/permissionmodel.cpp
```diff
@@ -103,8 +162,9 @@ void PermissionModel::remove(int index)
 
 void PermissionModel::setCapability(QModelIndex index, int capability)
```
llewelld, line 163: Is there any reason not to use `PermissionManager::Capability` here for the second parameter, instead of `int`?

import/controls/permissionmanager.cpp
```diff
@@ -34,14 +29,6 @@ void PermissionManager::remove(const QString &host, const QString &type)
...
 
-void PermissionManager::requestUriPermissions(const QString &uri, QJSValue callback)
-{
-    if (!callback.isNull() && !callback.isUndefined() && callback.isCallable()) {
-        m_callback.reset(new QJSValue(callback));
-    }
-    sendRequest(QStringLiteral("get-all-for-uri"), uri);
-}
-
 void PermissionManager::add(const QString &host, const QString &type, int capability)
```
llewelld, line 32: Similarly here, any reason this can't use `PermissionManager::Capability`?

import/controls/permissionfilterproxymodel.cpp
```diff
@@ -36,8 +38,9 @@ void PermissionFilterProxyModel::setCapability(int currentIndex, int capability)
     QModelIndex sourceIndex = mapToSource(proxyIndex);
```
llewelld, line 38: And finally here also (use of `PermissionManager::Capability` instead of `int`).

llewelld (approved):
These changes look good to me. I have one suggestion to use `PermissionManager::Capability` instead of `int` for the three `setCapability()` methods, but this is a minor point, so I approve either way.

import/controls/permissionmodel.cpp
atatarov: I've got something like that
https://bugreports.qt.io/browse/QTBUG-19741
enum doesn't work for Q_INVOKABLE method's in other class than where it is. In some case's I've mistaken.. I will find the places, where it is possible and will fix these

PR #81, opened 2020-11-19
[sailfish-components-webview] Add removing illegal characters and spaces from destination path. Fixes JB#51954

paveltkv:


rainemak (approved):
LGTM

PR #82, opened 2020-11-23
[sailfish-components-webview] Replace Geolocation with Positioning in "Geolocation is disabled" notice. Fixes JB#52125

paveltkv:


rainemak (approved):
Thanks

PR #86, opened 2020-12-10
[sailfish-webview] Append add method to proxyModel and setCapability if host already exist. Contributes to JB#52422

atatarov:
If you want to add new exception to permission model, you need make sure that it doesn't
have for this host. Otherwise, change capability for this host.

### DIFF
--- a/import/controls/permissionmodel.cpp
+++ b/import/controls/permissionmodel.cpp
@@ -43,15 +43,22 @@ void PermissionModel::componentComplete()
 
 void PermissionModel::add(const QString &host, const QString &type, int capability)
 {
+    int it = 0;
+    for (const auto &perm : m_permissionList) {
+        if (perm.m_host == host && perm.m_type == type) {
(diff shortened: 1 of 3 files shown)


import/controls/permissionmodel.cpp
```diff
@@ -43,15 +43,22 @@ void PermissionModel::componentComplete()
 
 void PermissionModel::add(const QString &host, const QString &type, int capability)
 {
+    int it = 0;
+    for (const auto &perm : m_permissionList) {
+        if (perm.m_host == host) {
```
llewelld, line 48: Should this be checking the `type` too? Otherwise it's impossible to add different permissions for different types on the same host.

llewelld: Is a straight string equality enough to test for matching hosts here? (e.g. is it certain by this point that the transport, port, resource name etc. have been stripped?).

llewelld (approved):

PR #89, opened 2020-12-15
[sailfish-components-webview] Limit length of downloaded file name to 255 bytes. Fixes JB#52518

paveltkv:


llewelld (changes requested):
This all looks good. Could you please just amend the commit message to mention something about multibyte characters (otherwise it looks like this is a change to limit the filename length, whereas in practice this was already happening, but without taking into account multibyte chars).

llewelld:
Nice UTF8 truncation approach by the way.

llewelld (approved):
Thanks; that's much clearer now. The code works nicely and the unit tests are also good, so LGTM.

PR #92, opened 2021-02-23
[sailfish-webview] Add translation generation script. Fixes JB#53252

Tomin1:
Add translation generation script for property files. This is intended
to be used when engine is updated and the result can be then committed
to a repository. The script outputs a C++ source file.

### DIFF
--- a/tools/generate_property_translations.py
+++ b/tools/generate_property_translations.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+#
+# Generates translation strings from gecko property files for lupdate
+# to process
+
+import os, os.path, sys, re, dataclasses, functools, argparse
+
+DEFAULT_PREFIX = "sailfish_components_webview_popups-la"
+
+class Generator:
+    TRANSLATION = re.compile(r"(?P<name>[\w-]+)\s*=\s*(?P<text>.*)\s*")
+    SPLITTER = re.compile(r"([A-Za-z0-9][a-z0-9]*)")
+    PLACEHOLDER = re.compile(r"%(?:([0-9]+)\$)?S")
+
+    def __init__(self, file, source):
+        self.file = file
+        self.source = source
+
+    @staticmethod
+    def _convert_key(name):
+        """Convert from camelCase to snake_case"""
+        return "_".join(Generator.SPLITTER.findall(name)).lower()
+
+    @staticmethod
+    def _convert_text(text):
+        """Replace gecko's placeholders with Qt's and
+        remove access key hints"""
+        def replacer(match):
+            return "%1" if match.group(1) is None else "%" + match.group(1)
+        new = Generator.PLACEHOLDER.sub(replacer, text)
+        while new != text:
+            text = new
+            new = Generator.PLACEHOLDER.sub(replacer, text)
+        return new.replace("&", "")
+
+    def process_line(self, line, lineno):
+        """Processes one line of input and returns Translation objects"""
+        match = self.TRANSLATION.match(line)
+        if match is not None:
+            return Translation(self.source, lineno, match.group('name'),
+                               self._convert_key(match.group('name')),
+                               self._convert_text(match.group('text')))
+
+    def __iter__(self):
+        try:
+            for lineno, line in enumerate(self.file, 1):
+                result = self.process_line(line, lineno)
+                if result is not None:
+                    yield result
+        except UnicodeDecodeError:
+            print(f"Invalid character in file: {self.file.name}",
+                  file=sys.stderr)
+            raise
+
+@dataclasses.dataclass
+class Translation:
+    source: str
+    lineno: int
+    original_key: str
+    key: str
+    text: str
+
+    prefix = None
+
+    def __str__(self):
+        return f"""
+//: This is a transformed translation key, the original translation key
+//: ({self.original_key}) is from {self.source} at line {self.lineno}
+//% "{self.text}"
+const auto {self.key} = qtTrId("{self.prefix}-{self.key}");"""
+
+def generate(file, url):
+    print("/* Generated dummy file, do not edit */")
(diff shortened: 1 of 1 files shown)


Tomin1:
> This is intended to be used when engine is updated and the result can be then committed to this repository.

This repository or a repository (some repository)?

rainemak:
Lines like:
neverForSiteButtonText = Ne&ver for This Site

Will result something like:

//% "Ne&ver for This Site"
const auto passwordmgr_never_for_site_button_text = qtTrId("sailfish_components_webview_popups-la-passwordmgr_never_for_site_button_text");

Those kind of cases look like keyboard shortcuts => as we don't have support either => let's just drop those. 

Nonetheless, already quite nice.

tools/generate_property_translations.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @staticmethod
+    def _convert_key(name):
+        """Convert from camelCase to snake_case"""
+        return "_".join(Generator.SPLITTER.findall(name)).lower()
+
+    @staticmethod
+    def _convert_text(text):
+        """Replace gecko's placeholders with Qt's and
+        remove keyboard shortcut marks"""
```
Tomin1, line 33: I guess hints would be more correct than marks.

rainemak: Yeap.

tools/generate_property_translations.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+#
+# Generates translation strings from gecko property files for lupdate
+# to process
+
+import os, os.path, sys, re, dataclasses, functools
+
+PREFIX = "sailfish_components_webview_popups-la"
```
rainemak, line 8: This could be an argument then with that this would be easier to utilize for other cases as well.

tools/generate_property_translations.py
```diff
@@ -0,0 +1,86 @@
...
+            raise
+
+@dataclasses.dataclass
+class Translation:
+    name: str
+    key: str
+    text: str
+
+    def __str__(self):
+        return f"""
```
rainemak, line 67: Hmm... wondering if we should actually have the original key here...

//: The original translation key - {self.original_key}

tools/generate_property_translations.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+#
+# Generates translation strings from gecko property files for lupdate
+# to process
+
+import os, os.path, sys, re, dataclasses, functools, argparse
+
+DEFAULT_PREFIX = "sailfish_components_webview_popups-la"
```
rainemak, line 8: Maybe cli argument ... as this is kind of translation key prefix.

tools/generate_property_translations.py
Tomin1: I added proper argument parsing and --prefix|-p for this one. Defaults to the old value.

tools/generate_property_translations.py
Tomin1: Added this too.

tools/generate_property_translations.py
Tomin1: Is this --prefix or -p somehow insufficient?

rainemak: Yes

tools/generate_property_translations.py
```diff
@@ -0,0 +1,97 @@
...
+    name: str
+    original_key: str
+    key: str
+    text: str
+
+    prefix = None
+
+    def __str__(self):
+        return f"""
+//: The original translation key - {self.original_key}
```
rainemak, line 71: Small tweak for this would be in place...

//: This is transformed translation key - the original translation key is {self.original_key} from file {self.source}

If source could be passed separately then script user could indicate nicely for example https://git.sailfishos.org/mer-core/gecko-dev/blob/master/toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties

Tomin1: I wonder if it would be useful to use urls instead of file names and then download the sources to build these files.

Tomin1: It wasn't quite so small tweak but now it's possible to have those urls as well. For example:
generate_property_translations.py --url https://git.sailfishos.org/mer-core/gecko-dev/blob/master/ toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties toolkit/locales/en-US/chrome/global/filepicker.properties
generates proper urls to the files but it means that this must be run on top level in that repository. Also version can be defined similarly.

rainemak: Seeing something like this as an output without version that is:

```
//: This is a transformed translation key,
//: the original translation key (insecureFieldWarningDescription3) is from https://git.sailfishos.org/mer-core/gecko-dev/blob/master/toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties../gecko-dev/toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties (unknown)
```

As I already gave the url I'd not expect to see that local file path. Ran like:

```
./tools/generate_property_translations.py  --url https://git.sailfishos.org/mer-core/gecko-dev/blob/master/toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties ../gecko-dev/toolkit/locales/en-US/chrome/passwordmgr/passwordmgr.properties
```

Tomin1: Well, the idea was that the translation file could be generated from more than one property file, thus the argument for --url needs to be common for all files. If that's not needed I can drop that feature and change it to be used like you described.

Tomin1: Changed this as discussed yesterday. Removed --version argument and support for multiple input files. Also translation ids no longer have file name in them.

Tomin1:
Added line numbers.

rainemak (approved):

PR #94, opened 2021-03-03
[sailfish-webview] Start using user agent overrides. Fixes JB#53103

mkenttala:


lib/webenginesettings.cpp
```diff
@@ -129,6 +129,10 @@ void SailfishOS::WebEngineSettings::initialize()
     engineSettings->setPreference(QStringLiteral("browser.enable_automatic_image_resizing"),
                                   QVariant::fromValue<bool>(true));
 
+    // Enable user agent overrides
```
llewelld, line 132: Copyright header needs refreshing on this file too.

import/webview/plugin.cpp
```diff
@@ -74,6 +79,33 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
...
 
+void SailfishOSWebViewPlugin::initUserAgentOverrides(const QString &path)
+{
+    QString mozillaDir = QString("%1/.mozilla/").arg(path);
+    QDir dir(mozillaDir);
+    if (!dir.exists())
+        dir.mkpath(dir.path());
+
+    QFile ua(mozillaDir + MOZILLA_DATA_UA_UPDATE);
+    if (!ua.exists()) {
```
llewelld, line 90: I may be misreading the logic here, but it looks like once the file is created, it will never be regenerated. Is this correct behaviour? It seems that if `ua-update.json.in` changes then this file should be regenerated too.

lib/webenginesettings.cpp
```diff
@@ -129,6 +129,10 @@ void SailfishOS::WebEngineSettings::initialize()
...
+    engineSettings->setPreference(QStringLiteral("general.useragent.updates.enabled"),
+                                  QVariant::fromValue<bool>(true));
```
llewelld, line 134: Should this be following `MGConfItem(QStringLiteral("/apps/sailfish-browser/settings/useragent_update_enabled"))`, or is that old behaviour now?

lib/webenginesettings.cpp
llewelld, line 132: I guess this needs the copyright header refreshing too.

llewelld (changes requested):
Tested and this does the trick nicely, so a really nice fix.

lib/webenginesettings.cpp
llewelld: (accidental duplicate)

import/webview/plugin.cpp
mkenttala: New version which adds missing overrides from ua-update.json.in.

lib/webenginesettings.cpp
mkenttala: I debugged that it's mandatory to have this set or the saved overrides are not read at all. Also this lib could be used by any app so using settings from sailfish-browser is a bit questionable.

rubdos:
I can confirm that this seem to fix the infamous Whisperfish reCAPTCHA bug.

lib/webenginesettings.cpp
llewelld: Yes, fair enough. It might be good to allow apps to set their own overrides, but that's something for the future.

import/webview/plugin.cpp
```diff
@@ -74,6 +83,68 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
     webEngine->addObserver(QStringLiteral("clipboard:setdata"));
 }
 
+void SailfishOSWebViewPlugin::initUserAgentOverrides(const QString &path)
```
llewelld, line 86: This could be a static function (it doesn't access any class members from what I can tell), but it's a small point; your choice whether to change or not.

import/webview/plugin.cpp
```diff
@@ -74,6 +83,68 @@ void SailfishOSWebViewPlugin::initializeEngine(QQmlEngine *engine, const char *u
...
+        }
+        QJsonObject source = sourceDoc.object();
+        QJsonObject dest = destDoc.object();
+        QStringList destKeys = dest.keys();
+        bool changed = false;
+        for (const QString &key : source.keys()) {
+            if (!destKeys.contains(key)) {
+                // Append missing key and value
+                dest.insert(key, source.value(key));
+                changed = true;
```
llewelld, line 134: Won't it still be the case that if a value is changed in `/usr/share/sailfish-browser/data/ua-update.json.in`, that it will be left unchanged by this?

Presumably the two should be kept synchronised at all times, but correct me if I'm misunderstanding. Originally I was thinking you could maybe just compare timestamps and refresh the entire file if ua-update.json.in was newer, but the approach you've used here should work too if it checks for value differences (rather than just existence of a key), I think.

mkenttala: I don't know if webview downloads these overrides automatically from net like sailfish-browser does, so I didn't want to overwrite already existing values with possibly older data. Anyway safest bet is to use ua-update.json.in values so I'll change this.

llewelld: I hadn't thought of that. This seems to update nicely now at any rate.

llewelld (approved):
This works nicely for me, thanks Marko. It may be worth checking back on the conversation [here](https://forum.sailfishos.org/t/whisperfish-the-unofficial-sailfishos-signal-client/3337/163) to ensure the issue experienced by ninepine is just an upgrade glitch.

PR #95, opened 2021-03-11
[components-webview] Fix saving file name with url escaping. Fixes JB#53509

pvuorela:
Saving file with spaces results in %20 in url. The handling here
didn't decode, but then removed % on downloadhelper.
-> saved file had spaces replaced with '20'.

Now the saved file doesn't still get spaces as the download helper
replaces whitespace with underscores, but that's a lot better already.

@rainemak


pvuorela:
Though pondering should those spaces also end up in the saved file. Those are anyway legal characters.

rainemak (approved):

PR #96, opened 2021-03-20
[sailfish-components-webview] Add ability to set permission for current session. Contributes to JB#52731

atatarov:
Now we have popup notifications, we need to have ability to set permission for the session.
Popup differs from other permissions (ex. geolocation) we don't recieve a request and can't handle it with
checkedDontAsk response. We have to catch popup blocked events ourselves and handle it.
Therefore, adding the capabilities of the PermissionManager is required.


llewelld (approved):
This looks good to me.

okodron (approved):

PR #98, opened 2021-04-15
[sailfish-webview] Implement WebRTC permission dialog. JB#53947

d-grigorev:
….

### DIFF
--- a/import/popups/PopupOpener.qml
+++ b/import/popups/PopupOpener.qml
@@ -136,11 +136,7 @@ Timer {
         }
         var rejectFn = acceptFn
 
-        var topic = "embed:alert"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:alert", null, props, acceptFn, rejectFn)
     }
 
     // Open confirm dialog
@@ -169,11 +165,7 @@ Timer {
             })
         }
 
-        var topic = "embed:confirm"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:confirm", null, props, acceptFn, rejectFn)
     }
 
     // Open prompt dialog
@@ -204,11 +196,7 @@ Timer {
             })
         }
 
-        var topic = "embed:prompt"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:prompt", null, props, acceptFn, rejectFn)
     }
 
     // Open login dialog
@@ -234,11 +222,7 @@ Timer {
             })
         }
 
-        var topic = "embed:login"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:login", null, props, acceptFn, rejectFn)
     }
 
     // Open password manager dialog
@@ -288,11 +272,7 @@ Timer {
             })
         }
 
-        var topic = "embed:auth"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:auth", null, props, acceptFn, rejectFn)
     }
 
     // Open permissions dialog
@@ -320,27 +300,17 @@ Timer {
             })
         }
 
-        var topic = "embed:permissions"
-        var subtopic = data.title
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+        openPopupByTopic("embed:permissions", data.title, props, acceptFn, rejectFn)
     }
 
     function webrtc(data) {
-        var topic = "embed:webrtcrequest"
-        var subtopic = null
-        var comp =_resolveListenerComponent(topic, subtopic)
-        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
-        var props = {
-            "origin": data.origin,
-            "devices": data.devices
-        }
-
         // Promote only supported requests and only from an observable origin
         if (data.origin && ("camera" in data.devices ||
                             "microphone" in data.devices)) {
-            openPopup(comp, props, compIsDialog,
+            openPopupByTopic("embed:webrtcrequest", null, {
+                    "origin": data.origin,
+                    "devices": data.devices
+                },
                 // accept
                 function(popup) {
                     contentItem.sendAsyncMessage("embedui:webrtcresponse", {
@@ -423,6 +393,12 @@ Timer {
                 obj.rejected.connect(function() { rejectedFn(obj) })
             }
         }
+    }
+
+    function openPopupByTopic(topic, subtopic, properties, acceptedFn, rejectedFn) {
+        var comp =_resolveListenerComponent(topic, subtopic)
+        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
+        openPopup(comp, properties, compIsDialog, acceptedFn, rejectedFn)
     }
 
     function _openContextMenu(data) {
--- a/import/popups/PopupOpener.qml
+++ b/import/popups/PopupOpener.qml
@@ -32,6 +32,7 @@ Timer {
         "embed:login":          "passwordManagerPopup",
         "embed:auth":           "authPopup",
         "embed:permissions":  { "geolocation": "locationPermissionPopup" },
+        "embed:webrtcrequest":  "webrtcPermissionPopup"
     })
     readonly property var listeners: Object.keys(_messageTopicToPopupProviderPropertyMapping)
 
@@ -111,6 +112,7 @@ Timer {
             }
             break
         }
+        case "embed:webrtcrequest": webrtc(data);   break;
         }
         // If we end up here, message has been handled.
         return true
@@ -323,6 +325,49 @@ Timer {
         var comp =_resolveListenerComponent(topic, subtopic)
         var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
         openPopup(comp, props, compIsDialog, acceptFn, rejectFn)
+    }
+
+    function webrtc(data) {
+        var topic = "embed:webrtcrequest"
+        var subtopic = null
+        var comp =_resolveListenerComponent(topic, subtopic)
+        var compIsDialog = _resolveListenerComponentType(topic, subtopic) === "dialog"
+        var props = {
+            "origin": data.origin,
+            "devices": data.devices
+        }
+
+        // Promote only supported requests and only from an observable origin
+        if (data.origin && ("camera" in data.devices ||
+                            "microphone" in data.devices)) {
+            openPopup(comp, props, compIsDialog,
+                // accept
+                function(popup) {
+                    contentItem.sendAsyncMessage("embedui:webrtcresponse", {
+                        "allow": true,
+                        "checkedDontAsk": popup.rememberValue,
+                        "choices": popup.choices,
+                        "id": data.id
+                    })
+                },
+                // reject
+                function(popup) {
+                    contentItem.sendAsyncMessage("embedui:webrtcresponse", {
+                        "allow": false,
+                        "checkedDontAsk": popup.rememberValue,
+                        "choices": popup.choices,
+                        "id": data.id
+                    })
+                }
+            )
+        } else {
+            contentItem.sendAsyncMessage("embedui:webrtcresponse", {
+                "allow": false,
+                "checkedDontAsk": false,
+                "choices": {},
+                "id": data.id
+            })
+        }
     }
 
     // Handle pagestack busy change
--- a/import/popups/WebrtcPermissionDialog.qml
+++ b/import/popups/WebrtcPermissionDialog.qml
@@ -1,0 +1,167 @@
+/****************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+UserPromptDialog {
+    id: dialog
+
+    property alias devices: webrtc.devices
+    property alias origin: webrtc.origin
+    property alias rememberValue: webrtc.rememberValue
+    property alias choices: webrtc.choices
+
+    WebrtcPermissionInterface {
+        id: webrtc
+
+        anchors.fill: parent
+        rememberValue: remember.checked
+
+        //: Allow the server to use media devices
+        //% "Allow"
+        acceptText: qsTrId("sailfish_components_webview_popups-he-allow_media_devices")
+
+        //: Deny the server access to media devices
+        //% "Deny"
+        cancelText: qsTrId("sailfish_components_webview_popups-he-deny_media_devices")
+
+        readonly property var dictionary: {
+            //% "Front camera"
+            "secondary": qsTrId("sailfish_components_webview_popups-la-primary_camera"),
+            //% "Rear camera"
+            "primary": qsTrId("sailfish_components_webview_popups-la-secondary_camera"),
+            //% "Integrated microphone"
+            "Integrated microphone": qsTrId("sailfish_components_webview_popups-la-integrated_microphone")
+        }
+
+        function translate(word) {
+            var translated = dictionary[word]
+            return translated ? translated : word
+        }
+
+        UserPromptUi {
+            anchors.fill: parent
+            dialog: dialog
+            popupInterface: webrtc
+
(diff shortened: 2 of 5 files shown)


atatarov:
It's look good, but there are two small questions:
1. I see the same code in the MicrophoneDialog.qml and LocationDialog.qml. Could it be as a PermissionDialog.qml or like that?
In this case, I will be able to use it for https://github.com/sailfishos/sailfish-browser/pull/815/files PopupBlockedDialog in sailfish-browser
2. 'webrtc:request' naming is fine, but there are the 'embed:' suffix for the same things, or 'embed:perms

@d-grigorev @llewelld

rainemak:
> It's look good, but there are two small questions:
> 
>     1. I see the same code in the MicrophoneDialog.qml and LocationDialog.qml. Could it be as a PermissionDialog.qml or like that?

@chriadam started looking API for hybrid app developers to override prompts. I think we should align with that instead. Hence, for time being it's not necessarily bad idea to have a bit of duplicated code.

>        In this case, I will be able to use it for https://github.com/sailfishos/sailfish-browser/pull/815/files PopupBlockedDialog in sailfish-browser
> 
>     2. 'webrtc:request' naming is fine, but there are the 'embed:' suffix for the same things, or 'embed:perms
> 
> 
> @d-grigorev @llewelld

Let's use similar message naming scheme and try to avoid introducing new ones. For instance, it could be like "embed:webrtcrequst". As @atatarov mentioned, those "embed" and "embedui" are for embedlite-components messages. In case you're wondering about case like "Content:ContextMenu", those are originating from similar Gecko components (usually Android FF components) that we have -- some older and than others. This way, it's simpler to distinguish which we have adapted from Gecko and which are our own implementations.

dgreenday:
I fixed message naming as you suggested.
I was planned to extend MicrophoneDialog.qml with device selection UI in the near future, so it will differ from LocationDialog.qml.

atatarov (approved):
It depends on 
https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/120

okodron (approved):

import/popups/PopupOpener.qml
```diff
@@ -184,6 +185,26 @@ Timer {
             root._openContextMenu(data)
             break
         }
+        case "embed:webrtcrequest": {
+            if (data.title === "microphone") {
+                var obj = pageStack.animatorPush(Qt.resolvedUrl("MicrophoneDialog.qml"), {"host": data.host })
```
llewelld, line 190: Using the test page, if I click the "Microphone" button in quick succession it can trigger two `embed:webrtcrequest` requests before the page has had time to push. It results in the following error:
```
JSComp: EmbedLiteWebrtcUI.js: Created request camera:{02190832-072d-4d90-b994-8bea23633cf1}
[D] doPush:134 - Warning: cannot push while transition is in progress
[W] unknown:191 - file:///usr/lib/qt5/qml/Sailfish/WebView/Popups/PopupOpener.qml:191: TypeError: Cannot read property 'pageCompleted' of null
JSComp: EmbedLiteWebrtcUI.js: Allow video device -1
```
This causes a bit of a mess because it means `embedui:webrtcresponse` is never sent, so `WebrtcPermissionRequest` is never resolved, and so most importantly `Services.embedlite.removeMessageListener("embedui:webrtcresponse", this)` is never called in `EmbedLiteWebrtcUI._prompt()`. Future calls therefore result in multiple calls to `EmbedLiteWebrtcUI.onMessageReceived()` causing an error. See the following output:

```
JSComp: EmbedLiteWebrtcUI.js: Audio devies 6 video devices 0
JSComp: EmbedLiteWebrtcUI.js: Created request microphone:{481c2e96-ea9c-4bd2-9809-2d5985fe9df5}
JSComp: EmbedLiteWebrtcUI.js: Submitting request microphone id microphone:{481c2e96-ea9c-4bd2-9809-2d5985fe9df5}
JSComp: EmbedLiteWebrtcUI.js: Created request camera:{481c2e96-ea9c-4bd2-9809-2d5985fe9df5}
JSComp: EmbedLiteWebrtcUI.js: Allow video device -1
JSComp: EmbedLiteWebrtcUI.js: onMessageReceived: top:embedui:webrtcresponse, msg:{"allow":true,"checkedDontAsk":false,"id":"microphone:{481c2e96-ea9c-4bd2-9809-2d5985fe9df5}"}
JSComp: EmbedLiteWebrtcUI.js: onMessageReceived: top:embedui:webrtcresponse, msg:{"allow":true,"checkedDontAsk":false,"id":"microphone:{481c2e96-ea9c-4bd2-9809-2d5985fe9df5}"}
JSComp: EmbedLiteWebrtcUI.js: Allow audio device 0
JSComp: EmbedLiteWebrtcUI.js: Selected audio device: default: Null Input
JSComp: EmbedLiteWebrtcUI.js: getUserMedia:response:allow for callID {481c2e96-ea9c-4bd2-9809-2d5985fe9df5}
JSComp: EmbedLiteWebrtcUI.js: got getUserMedia:request
JavaScript error: file:///usr/lib/mozembedlite/components/EmbedLiteWebrtcUI.js, line 225: NS_ERROR_UNEXPECTED: 
CONSOLE message:
[JavaScript Error: "NS_ERROR_UNEXPECTED: " {file: "file:///usr/lib/mozembedlite/components/EmbedLiteWebrtcUI.js" line: 225}]
```

llewelld (changes requested):
This is really neat. I have a few small requests and also some suggestions.

import/popups/PopupOpener.qml
llewelld: Just an important addendum: I added some extra lines to the code for debug output, so the line number for that javascript is 219 in [this version](https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/120/diffs) of the code.

d-grigorev: Thank you for review. The problem with pagestack transition is fixed.  I rewrote the dialog to support cameras as well. This depends on the new version of https://git.sailfishos.org/mer-core/embedlite-components/merge_requests/120.

import/popups/MicrophoneDialog.qml
```diff
@@ -0,0 +1,55 @@
...
+
+    property string host
+
+    property alias rememberValue: remember.checked
+
+    //: Allow the server to use the microphone
+    //% "Allow"
+    acceptText: qsTrId("sailfish_components_webview_popups-he-accept_microphone")
+
+    //: Deny the server to use the microphone
```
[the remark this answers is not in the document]
d-grigorev: Fixed. Thanks.

import/popups/MicrophoneDialog.qml
```diff
@@ -0,0 +1,55 @@
+/****************************************************************************
+**
+** Copyright (c) 2013 - 2016 Jolla Ltd.
+** Copyright (c) 2020 - 2021 Open Mobile Platform LLC.
```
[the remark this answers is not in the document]
d-grigorev: Initially the file was copied from the neighbor LocationDIalog.qml, it has Jolla copyright. But since then, the content has completely changed, so I left only OMP's. Thanks!

import/popups/PopupOpener.qml
```diff
@@ -184,6 +185,26 @@ Timer {
...
+                    dialog.accepted.connect(function() {
+                        contentItem.sendAsyncMessage("embedui:webrtcresponse",
+                                                     { "allow": true, "checkedDontAsk": dialog.rememberValue, "id": data.id })
+                    })
+                    dialog.rejected.connect(function() {
+                        contentItem.sendAsyncMessage("embedui:webrtcresponse",
+                                                     { "allow": false, "checkedDontAsk": dialog.rememberValue, "id": data.id })
+                    })
+                })
+            } else {
```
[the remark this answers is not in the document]
d-grigorev: Done.

import/popups/PopupOpener.qml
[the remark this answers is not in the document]
d-grigorev: Done.

import/popups/PopupOpener.qml
```diff
@@ -184,6 +185,26 @@ Timer {
             root._openContextMenu(data)
             break
         }
+        case "embed:webrtcrequest": {
+            if (data.title === "microphone") {
+                var obj = pageStack.animatorPush(Qt.resolvedUrl("MicrophoneDialog.qml"), {"host": data.host })
+                obj.pageCompleted.connect(function(dialog) {
+                    dialog.accepted.connect(function() {
+                        contentItem.sendAsyncMessage("embedui:webrtcresponse",
+                                                     { "allow": true, "checkedDontAsk": dialog.rememberValue, "id": data.id })
```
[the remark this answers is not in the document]
d-grigorev: The logic has been changed, now device selection and cameras are supported.

import/popups/PopupOpener.qml
```diff
@@ -82,47 +98,18 @@ Timer {
...
         case "embed:alert": {
-            if (pageStack.busy) {
-                // Wait until pagestack ready
-                root.data = data
-                root.topic = topic;
-                pageStack.busyChanged.connect(busyChanged)
-            } else {
-                alert(data)
-            }
+            popupOpen(alert, data)
```
llewelld, line 101: It's just a thought, but since `data` doesn't change, you could even simplify the logic by binding the data directly to the function at this point (something like `popupOpen(alert.bind(data))`, but this is untested).

llewelld (commented):
These look like really nice and solid changes. I've made some suggestions, but they're all just ideas.

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,159 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+UserPrompt {
```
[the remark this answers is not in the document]
d-grigorev: I need some time to rework changes in `PopupOpener.qml`. It is becoming too complex on my opinion. Don't you think I should create another popup class for webrtc and make it customizable as well?

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,159 @@
...
+        "secondary": qsTrId("sailfish_components_webview_popups-la-primary_camera"),
+        //% "Rear camera"
+        "primary": qsTrId("sailfish_components_webview_popups-la-secondary_camera"),
+        //% "Integrated microphone"
+        "Integrated microphone": qsTrId("sailfish_components_webview_popups-la-integrated_microphone")
+    }
+
+    function translate(word) {
+        var translated = dictionary[word]
+        return translated ? translated : word
```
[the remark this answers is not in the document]
d-grigorev: I may be wrong, but I don't think `word` should be capitalized. It is an ID of media hardware from gecko. In general case it can contain, for example the model name of UVC camera or something and will be less readable if capitalized, I believe. It would be nice to be able to make the dictionary hardware dependent, but I don't know how.

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,159 @@
...
+
+    Component.onCompleted: {
+        if ("microphone" in devices && "camera" in devices) {
+            choices = {"microphone": 0, "camera": 0}
+            //% "Camera and microphone"
+            title.text = qsTrId("sailfish_components_webview_popups-la-webrtc_camera_microphone_title")
+            //: %1 is the site that wants to use camera and microphone
+            //% "Allow %1 to use your camera and microphone?"
+            label.text = qsTrId("sailfish_components_webview_popups-la-camera_microphone_request")
+                                .arg(host)
```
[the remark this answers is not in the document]
d-grigorev: This will mislead about permanent permissions.  I'm thinking about `host: this.uri.host ? this.uri.host : this.uri.filePath` on the embedlite side. The request should be rejected if embedlite fails to provide something readable here, what do you think?

import/popups/PopupOpener.qml
d-grigorev: I dropped this commit. Since customizable popus was merged, this change is unnecessary.

import/popups/WebrtcPermissionDialog.qml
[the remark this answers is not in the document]
llewelld: Okay, fair enough, I was expecting something like `microphone` or `camera`. In this case you are right, better to leave them as they are.

import/popups/WebrtcPermissionDialog.qml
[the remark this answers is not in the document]
llewelld: What is the effect of making a permanent exception for an empty host in practice? It would allow for more access than just the single file I think, or am I wrong about that? Essentially, I agree the wording needs to reflect exactly the permissions being granted.

I also agree, if there's something odd about the request then better to just reject it. However, granting access for use by a locally stored file sounds like something we should probably support.

import/popups/PopupOpener.qml
llewelld: Ah, okay. Is there a replacement commit?

d-grigorev: Yes. I that's it.

import/popups/WebrtcPermissionDialog.qml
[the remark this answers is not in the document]
d-grigorev: The permanent permission is stored with an uri, no matter, file or host in the web. So if you give the permission to `webrtc.html`, it is only valid for this particular file. However, (I'm thinking loud) what if the file has changed since the permission was granted? Perhaps we should not give permanent permissions to files unless they come from a secure context.

import/popups/WebrtcPermissionDialog.qml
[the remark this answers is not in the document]
d-grigorev: I've updated the PR to use this new customization feature.

import/popups/PopupOpener.qml
d-grigorev: I couldn't resist improving `PopupOpener.qml` a bit in the last commit.

import/popups/PopupOpener.qml
```diff
@@ -32,6 +32,7 @@ Timer {
         "embed:login":          "passwordManagerPopup",
         "embed:auth":           "authPopup",
         "embed:permissions":  { "geolocation": "locationPermissionPopup" },
+	"embed:webrtcrequest":  "webrtcPermissionPopup"
```
chriadam, line 35: nitpick: use spaces rather than tabs for consistency

import/popups/PopupOpener.qml
```diff
@@ -380,6 +395,12 @@ Timer {
         }
     }
 
+    function openPopupByTopic(topic, subtopic, properties, acceptedFn, rejectedFn) {
```
chriadam, line 398: nice idea.  I should have done this originally to reduce boilerplate.  thanks.

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,174 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+UserPromptDialog {
```
chriadam, line 14: This looks good to me.

Open question: what is your opinion of the extension mechanism / popup interfaces?  Are there pain points / things we should change now, before we ship?  Or was it fairly straightforward to use, in your opinion?

import/popups/qmldir
```diff
@@ -24,3 +24,5 @@ PromptLabel 1.0 PromptLabel.qml
 DownloadMenuItem 1.0 DownloadMenuItem.qml
 ShareLinkPage 1.0 ShareLinkPage.qml
 ShareTextPage 1.0 ShareTextPage.qml
+WebrtcPermissionInterface 1.0 WebrtcPermissionInterface.qml
```
chriadam, line 27: nitpick: I'd prefer to group the Interface types together, and the Dialog types together in the qmldir, just for readability

chriadam: i.e. move line 27 to 12/31, and move line 28 to 20/21.

chriadam (approved):
Very minor nitpicks but LGTM

import/popups/qmldir
d-grigorev: Done. Thanks.

import/popups/WebrtcPermissionDialog.qml
d-grigorev: At first glance, this is a little confusing, but after a thought, I can't figure out what could be improved.

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,174 @@
...
+                    width: parent.width - x * 2
+                    horizontalAlignment: Text.AlignHCenter
+                    font.pixelSize: Theme.fontSizeSmall
+                    color: Theme.highlightColor
+                }
+
+                ComboBox {
+                    visible: "camera" in devices
+                    x: Theme.horizontalPageMargin
+                    width: parent.width - x * 2
```
llewelld, line 98: ComboBoxes have an implicit margin and default to the parent width, so these can be skipped.
```suggestion
```

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,174 @@
...
+                            }
+                        }
+                    }
+                    onCurrentIndexChanged: choices["camera"] = currentIndex
+                }
+
+                ComboBox {
+                    visible: "microphone" in devices
+                    x: Theme.horizontalPageMargin
+                    width: parent.width - x * 2
```
llewelld, line 117: Similarly, better to remove these to get the standard margins.
```suggestion
```

import/popups/WebrtcPermissionDialog.qml
```diff
@@ -0,0 +1,174 @@
...
+UserPromptDialog {
+    id: dialog
+
+    property alias devices: webrtc.devices
+    property alias origin: webrtc.origin
+    property alias rememberValue: webrtc.rememberValue
+    property alias choices: webrtc.choices
+
+    property alias acceptText: webrtc.acceptText
+    property alias cancelText: webrtc.cancelText
```
llewelld, line 23: Are the `acceptText` and `cancelText` aliases needed, or can they be dropped? I didn't see them being used.

llewelld (changes requested):
This is good and the code looks nice. Just a couple of margin suggestions.

import/popups/WebrtcPermissionDialog.qml
d-grigorev: I wasn't sure, I copied the same from LocationDialog.qml. It seems they are not needed at all.

import/popups/WebrtcPermissionDialog.qml
d-grigorev: Yes. I missed that small detail. Fixed.

llewelld (approved):
Thanks for making the few changes. This looks good to me now.

PR #100, opened 2021-04-16
[oauth] Implement OAuth-related library and QML components. Contributes to: OMP#JOLLA-58 JB#53788

chriadam:


chriadam:
@rainemak @llewelld @okodron obviously the naming / namespacing / etc is still to be discussed.  But in terms of API: is this moving in the right direction?  It includes both C++ and QML APIs.  Examples work (once you put in clientId+clientSecret / consumerKey+consumerSecret, if your app allows localhost redirect as per developer console on remote service).

Still need to do documentation and polish, and then check other APIs to see if this one lacks some important features, especially things which might be required for OpenID Connect, going forward.

chriadam:
Used pimpl/dptr pattern in C++ library to ensure that modifications can be made in BC manner if required.  Added documentation.  Other minor polish.  I think this can be considered mostly ready-for-review at this stage.

oauth/doc/sailfish-oauth-oauth10a.qdoc
```diff
@@ -0,0 +1,186 @@
+/****************************************************************************
+**
+** Copyright (C) 2021 Open Mobile Platform LLC
```
atatarov, line 3:
```suggestion
** Copyright (c) 2021 Open Mobile Platform LLC
```
please, same for others

oauth/lib/redirecturiparser.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <QtCore/QPair>
+#include <QtCore/QList>
+
+QVariantMap SailfishOS::OAuth::parseRedirectUri(const QString &redirectUri, QString *errorMessage)
+{
+    QVariantMap data;
+
+    const QUrl url(redirectUri);
+    const QUrlQuery query(url.query());
+    const QString fragment = url.fragment(QUrl::FullyEncoded);
```
atatarov, line 23: Where were it used? I don't find this place

oauth/lib/redirecturiparser.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const QStringList fragments = fragment.split(QChar('&'));
+
+    QString errorString;
+
+    QList<QPair<QString, QString> > fragmentPairs;
+    for (const QString &f : fragments) {
```
atatarov, line 29: Could you rename please 'f' to 'fragment' or something more clear? (if the line 23 is unused)

oauth/examples/sailfishoauthqmlexample/sailfishoauthqmlexample.qml
```diff
@@ -0,0 +1,116 @@
...
+            Rectangle {
+                id: bottomRect
+                width: topRect.width
+                height: topRect.height
+                y: topRect.height
+                color: "lightgreen"
+
+                MouseArea {
+                    anchors.fill: parent
+                    property bool triggered: false
```
atatarov, line 64:
```suggestion
                    property bool triggered
```
It is false by default

oauth/examples/sailfishoauthqmlexample/sailfishoauthqmlexample.qml
```diff
@@ -0,0 +1,116 @@
...
+
+            Rectangle {
+                id: topRect
+                width: parent.width
+                height: parent.height / 2
+                color: "lightsteelblue"
+
+                MouseArea {
+                    anchors.fill: parent
+                    property bool triggered: false
```
atatarov, line 19: I see 'WIP' status and I hope it will be used anywhere in the future

oauth/doc/sailfish-oauth-oauth10a.qdoc
chriadam: Done

oauth/lib/redirecturiparser.cpp
chriadam: In the line below, but not necessary to be separated, so I have combined those two lines now.  Done.

oauth/lib/redirecturiparser.cpp
chriadam: Done

oauth/examples/sailfishoauthqmlexample/sailfishoauthqmlexample.qml
chriadam: Done, and same for the triggered property of the other mouse area

oauth/examples/sailfishoauthqmlexample/sailfishoauthqmlexample.qml
chriadam: The triggered property itself?  It just exists to prevent the oauth flow from being started multiple times.

If you mean the example itself: well, yes, it's quite ugly at the moment, exists solely to test that the API works and be quite minimal.  Once we fix the naming etc it would be nice to spend more time making the example more user-friendly also (e.g. show timeline tweets or something, rather than just printing data to console).

gorgool:
Hello, @okodron asked to review the PR. Unfortunately, I am not familiar with OAuth1, but there are a few questions and suggestions for OAuth2 component. Some of the comments below partially rely on, and make sense, only with a slightly revised architecture, suggestions for which will also be further.

*RedirectListener component*

You have made RedirectListener a public component. Have you considered making it part of OAuth component implementation?
As I understand it, redirectUri is a set when you register a client on the authentication server and the user will set it using OAuth component API. It might be worth hide the RedirectListener and determining the port (also host and scheme - http, https) for it based on the given value.

Suggestion: add the ability to disable timeout (for example via setTimeout(-1))

*OAuth2 component*

It might be worth removing state and codeVerifier (for PKCE) from the API. They are too technical and when they needed
all users will set them in the same way. It seems that it is better to enable and disable them by setting some flag

+ Add ability to reset or cancel the already running auth process.

oauth/lib/oauth2.cpp
```diff
@@ -0,0 +1,987 @@
...
+        d->m_errorMessage = errorMessage;
+        emit errorMessageChanged();
+    }
+    return data;
+}
+
+/*!
+  \brief Joins the specified \a scopes with the specified \a separator and returns the result as a valid \c{scope} string
+*/
+QString SailfishOS::OAuth::OAuth2::generateScope(
```
gorgool, line 604: Maybe you can have setScope overload that you can pass QStringList and separator (with default value)

oauth/lib/oauth2.cpp
```diff
@@ -0,0 +1,987 @@
...
+        emit errorChanged();
+        emit errorMessageChanged();
+        return;
+    }
+
+    QUrlQuery query;
+    query.addQueryItem("grant_type", "refresh_token");
+    query.addQueryItem("refresh_token", refreshToken);
+    query.addQueryItem("client_id", d->m_clientId);
+    if (d->m_flowType == AuthorizationCodeFlow
```
gorgool, line 899: I think you can simplify the condition to check if the clientSecret is not empty. If the user sets it perhaps he want it to be used. Flow type you already checked above

oauth/lib/oauth2.cpp
```diff
@@ -0,0 +1,987 @@
...
+            emit errorChanged();
+            emit errorMessageChanged();
+        } else if (accessTokenRequest->error() != QNetworkReply::NoError) {
+            const int httpCode = accessTokenRequest->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+            const QByteArray replyData = accessTokenRequest->readAll();
+            d_ptr->m_error = HttpError;
+            d_ptr->m_errorMessage = QStringLiteral("HTTP error: %1: %2: %3")
+                                       .arg(httpCode)
+                                       .arg(accessTokenRequest->errorString())
+                                       .arg(QString::fromUtf8(replyData));
```
gorgool, line 969: For non-HTTP type errors, replyData would be empty. Formatting will be off a little.

oauth/lib/redirectlistener.cpp
```diff
@@ -0,0 +1,325 @@
...
+    if (d->m_port > 0 && d->m_port < 65534) {
+        // use the port specified by the client
+        const quint16 p = d->m_port;
+        if (d->m_tcpServer->listen(QHostAddress(QHostAddress::LocalHost), p)) {
+            d->m_uri = QStringLiteral("http://127.0.0.1:%1/").arg(p);
+            emit uriChanged();
+            found = true;
+        }
+    } else {
+        // search for an available port
```
gorgool, line 283: I think you should not do it. Port is a part of a contract, that you specify when you register a client

oauth/lib/redirecturiparser.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    QVariantMap data;
+
+    const QUrl url(redirectUri);
+    const QUrlQuery query(url.query());
+    const QStringList fragments = url.fragment(QUrl::FullyEncoded).split(QChar('&'));
+
+    QString errorString;
+
+    QList<QPair<QString, QString> > fragmentPairs;
```
gorgool, line 27: QUrlQuery has a method queryItems(), that return parsed QList<QPair<QString, QString>>. You don't need to parse it by yourself. Can't you use it here?

chriadam:
> Hello, @okodron asked to review the PR. Unfortunately, I am not familiar with OAuth1, but there are a few questions and suggestions for OAuth2 component. Some of the comments below partially rely on, and make sense, only with a slightly revised architecture, suggestions for which will also be further.

Thanks very much for reviewing!  Much appreciated.

> 
> _RedirectListener component_
> 
> You have made RedirectListener a public component. Have you considered making it part of OAuth component implementation?
> As I understand it, redirectUri is a set when you register a client on the authentication server and the user will set it using OAuth component API. It might be worth hide the RedirectListener and determining the port (also host and scheme - http, https) for it based on the given value.

I wanted to allow the client application developer to implement their own custom redirect handling if required, because for example we don't currently support custom-uri-redirects.  We still hide the redirect handler in the common case (see e.g. sailfishoauthqmlexample.qml and the use of the OAuth2AcPkce type - there's no hint of redirect handler required).

In general, I wanted to create an API which hides complexity as much as possible, but still allows the developer access to low level details if required (e.g. via the C++ API, or even via property lookup on the OAuth2AcPkce object to access the redirect listener directly, etc).  "Make the common case easy, make the advanced case possible" sort of thing.

> Suggestion: add the ability to disable timeout (for example via setTimeout(-1))

Good point, I will make -1 mean disabled, as you suggest.

> _OAuth2 component_
> 
> It might be worth removing state and codeVerifier (for PKCE) from the API. They are too technical and when they needed
> all users will set them in the same way. It seems that it is better to enable and disable them by setting some flag.

Again, we hide these details in the common case (see e.g. sailfishoauthqmlexample.qml and the use of the OAuth2AcPkce type - there's no hint of state or codeVerifier properties being required).

> tokenEndpoint and refreshEndpoint are the same, it might confuse user. Maybe its better to set only tokenEndpoint and point out in docs that it also used in refresh operation.

Always?  I had believed that there might be some services which expose refresh on a different endpoint, which is why I separated that one out.  I may have been wrong to do so?

> It might be helpful to be able to reset or cancel the already running auth requests.

Yes, good point, I should add cancel() etc.

> _Errors_
> 
> You separated the error type (enum) and error description (errorMessage) and created different signals for them. At the same time, wherever you set them, you always do it together. Perhaps it is better to use one signal, and let the caller get errorMessage on-demand after it received the error signal.

There is one case where we emit errorMessage but not error - it is during parsing, when it's possible to encounter a non-fatal error.  For use from QML context, all properties should have notify signal defined so that change handlers and property bindings will be re-evaluated when required.

> Also, unfortunately, for HTTP errors, an error code is needed, since 5** or 4 ** requires a different action from the user.
> Perhaps it is worth using a richer class for errors.

Are you able to provide more information about precisely what errors might be encountered, and what different actions might be required?  I guess these are errors which don't occur during the web-browser flow, but instead during code-for-token exchange step or similar?

> _Another suggestion_
> 
>     1. Current implementation of RedirectListener supports only http (plain QTcpSocket). Would be nice to add support of QSslSocket to support https as well.

I agree - and also custom URI handling eventually.  But both of these are out of scope for the moment, I think, as we don't have much time remaining to complete this task...

>     2. Add support of auth discovery endpoint:
>        RFC: https://datatracker.ietf.org/doc/html/rfc8414
>        Examples:
> 
> 
>     * https://connect2id.com/products/server/docs/api/discovery
> 
>     * https://developers.google.com/identity/protocols/oauth2/openid-connect#discovery.
>       Then, for example, it will be possible to set "dicoveryEndpoint" and "autoDiscover" to true which will cause "tokenEndpoint" и "authEndpoint" to be set automatically.

Definitely, this one would be great to have.  Again, out of scope for this initial implementation I think.

>     1. Add ability to set user defined NetworkManager (subclass of QNetworkAccessManager). It can be used to set custom http(s) headers, caching, and more fine-grained settings.

I will add support for this one, now, as I think that customisation point might be important, and should only take a small amount of time.

oauth/lib/oauth2.cpp
chriadam: I can add such overload also.  The current method exists to make it simpler to use from QML (see e.g. how the OAuth2AcPkce type offers a "scopes" property and then binds the "scope" property to it through a generateScope() filter).

oauth/lib/redirectlistener.cpp
chriadam: Do you mean, on the server side?  The few services I've interacted with haven't required the client application to define the port they listen on, if using a localhost redirect.

Or do you simply mean, it's kind of bad to just search for an available port, in general?  That's fair enough, I suppose.  I'd like to keep the ability for the user to not care about redirect at all, in the common case (e.g. using OAuth2AcPkce.qml type), so I wonder what the best way is (e.g. I'd prefer not to expose a "redirectPort" property at the top level, so better might be to just specify a default port to use, although this might cause issues if that port is used by something else or if multiple applications are running, etc)...

oauth/lib/redirecturiparser.cpp
chriadam: That queryItems() method is used at the end of the function, but it only allows parsing query items, and doesn't parse key=value pairs from fragments.  This custom parseRedirectUri() method exists to ensure that whether the server provides the result as query items or as encoded key/value pairs in a fragment, we handle both in identical way.

oauth/lib/oauth2.cpp
chriadam: Good point, will remove those unnecessary conditions

oauth/lib/oauth2.cpp
chriadam: Good point, I will add a branch for that case and produce two different error messages to avoid confusion to the developer in that case.

oauth/lib/oauth2.cpp
chriadam: done

oauth/lib/oauth2.cpp
chriadam: done

oauth/lib/oauth2.cpp
chriadam: done

chriadam:
Rebased and made a variety of changes as per review comments.  Still need to discuss further what to do with the RedirectListener's port etc.

oauth/lib/redirectlistener.cpp
gorgool: I meant that even if you find a free port and open a socket on it, the server will still make the last redirect to port that you specify when registered the client on the server. There may be a situation when the original port is busy, then you found another one, but no redirect from the browser or webview will come, because the server does not know about the port been changed. For example, the authentication service that we use will return an error if you pass (in first authentication request) it a redirect uri different from the registered one.
First point in https://www.ory.sh/hydra/docs/debug/
Maybe there are existing servers with more relaxed requirements, that allow changing redirect uri. In that case, perhaps you can add some kind of option (searchAvailablePort(true | false)) that allows to switch this behaviour.

oauth/lib/redirecturiparser.cpp
gorgool: Got it. Didn't take into account these uri fragments.

oauth/lib/redirectlistener.cpp
gorgool: I see. Google allows to change port
https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server

gorgool:
About RedirectListener
Perhaps you are right, and you can leave the implementation to the developer. It seemed to me that to assume
that an application developer will write a custom RedirectListener is not very good. And if you stick to the protocol
then the response processing will be the same everywhere - you will receive a code (+ state) or an error. Then you will exchange the code for a token.
If popular services do not follow the standard, then the custom RedirectListener approach is fully justified.

> Always? I had believed that there might be some services which expose refresh on a different endpoint, which is why I separated that one out. I may have been wrong to do so?
If we stick to the standard, then  OAuth2 server can only have 2 endpoints (section 3). And token_endpoint is responsible for refreshing the token (section 3.2)
https://datatracker.ietf.org/doc/html/rfc6749#section-3
https://datatracker.ietf.org/doc/html/rfc6749#section-3.2

> Are you able to provide more information about precisely what errors might be encountered, and what different actions might be required? I guess these are errors which don't occur during the web-browser flow, but instead during code-for-token exchange step or similar?
Yes, exactly when exchanging (or refreshing) a token. If the developer sets the wrong token_endpoint, then the server can return 404
Moreover, if the server returned 500, then the endpoints may be correct, but the server itself is temporarily unavailable.
Developer can see this through errorMessage, but if we have to make a decision based on HTTP code parsing the error string might be error prone

oauth/lib/redirectlistener.cpp
chriadam: Yeah.  I think having the ability to "automatically" choose a port is useful.  The client can set the port manually if required (e.g. `redirectListener.port: 12345`) in which case the RedirectListener will use that port only.  I've updated the OAuth2AcPkce documentation and example to set the port manually, as it's likely to be a common use-case.

chriadam:
After giving it some thought, I think you're right: we should provide a separate Error type which contains code, message, and httpCode.  I've updated the PR with this change now.  I was wrong to use the errorMessage as an out-of-band signal for non-fatal parse error case, so instead I've now added a separate signal for that one also.

gorgool:
I can not find where you reset the error to NoError. If I make 2 consecutive requests, the first of which fails (set Error::Something), and the second succeeds, then on the same OAuth instance, after the second request, error() will return an error to me.

chriadam:
Thanks, good catch.  Now fixed, by resetting the error when starting each operation.

gorgool:
I finished. I don't see anything else. No more niggling.

chriadam:
Thanks very much for your comprehensive review!  I've removed the WIP flag, and rebased on master.

chriadam:
Moved to https://github.com/sailfishos/amber-web-authorization/pull/1

PR #101, opened 2021-04-20
Add scaffolding for doc generation

llewelld:
Currently just adds some scaffolding to be populated with the documentation.

### DIFF
--- a/doc/index.qdoc
+++ b/doc/index.qdoc
@@ -1,0 +1,78 @@
+/****************************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC
+** All rights reserved.
+**
+****************************************************************************************/
+
+/*!
+\page index.html
+\title Sailfish WebView
+
+The Sailfish WebView provides a straightforward and flexible way for you
+to introduce access to Web pages and websites into your applciation.
+
+The WebView component utilises the Sailfish OS browser engine, built using
+Mozilla gecko.
+
+\section1 Using Sailfish WebView
+
+Currently you have to explicitly link against qtmozembed-qt5 at build time, to
+avoid a symbol issue which leads to a crash:
+
+\code
+CONFIG += link_pkgconfig
+PKGCONFIG += qt5embedwidget
+\endcode
+
+\section2 C++ API
+
+The C++ API consists of the following classes:
+
+\list
+\li the \l{SailfishOS::WebEngine}{WebEngine} class
+\li the \l{SailfishOS::WebEngineSettings}{WebEngineSettings} class
+\endlist
+
+\section2 QML API
+
+The QML API is provided through the Sailfish.WebView import namespace
+
+The QML types exposed in the QML API are as follows:
+
+\list
+\li \l{WebView}
+\li \l{WebViewFlickable}
+\li \l{WebViewPage}
+\li \l{RawWebView}
+\endlist
+
+\section1 Example
+
+The following examples demonstrates how the WebView component can be integrated into an existing application's QML code.
+
+Example:
+
+\code
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+import Sailfish.WebView 1.0
+
+ApplicationWindow {
+    id: root
+    initialPage: Component {
+        Page {
+            id: page
+            WebView {
+                anchors.fill: parent
+                active: true
+                url: "http://www.sailfishos.org"
+            }
+        }
+    }
+}
+\endcode
+
+
+
+*/
--- a/doc/sailfish-webview.cpp
+++ b/doc/sailfish-webview.cpp
@@ -1,0 +1,31 @@
+namespace SailfishOS {
+namespace WebView {
+}
+}
+
+/*!
+  \module SailfishWebView
+*/
+
+/*!
+  \namespace SailfishOS
+  \inmodule SailfishWebView
+  \brief This namespace provides various APIs to interact with the Sailfish OS system.
+*/
+
+using namespace SailfishOS;
+
+/*!
+  \namespace WebView
+  \inmodule SailfishWebView
+  \brief This namespace provides components for displaying in-app web content.
+*/
+
+/*!
+\qmlmodule Sailfish.WebEngine 1.0
+\title Sailfish.WebEngine QML Types
+This module includes the following types:
+\generatelist{qmltypesbymodule Sailfish.WebEngine}
+*/
+
+using namespace SailfishOS::WebView;
--- a/doc/sailfish-webview.qdocconf
+++ b/doc/sailfish-webview.qdocconf
@@ -1,0 +1,56 @@
+project         = Sailfish WebView
+description     = Sailfish Webview Reference Documentation
+versionsym      =
+version         = 1.0.0
+url             = https://github.com/sailfishos/sailfish-webview/
+
+sourcedirs += $$PWD/../lib $$PWD/../import $$PWD/../doc
+headerdirs += $$PWD/../lib $$PWD/../import
+#exampledirs += $$PWD/../examples
+
+outputformats = HTML
+outputdir = $$PWD/../doc/html
+base = file:$$PWD/../doc/html
+tagfile = $$PWD/../doc/html/sailfishos-webview.tags
+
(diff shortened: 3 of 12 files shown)


doc/sailfish-webview.cpp
```diff
@@ -0,0 +1,31 @@
+namespace SailfishOS {
+namespace WebView {
+}
+}
+
+/*!
+  \module SailfishOAuth
```
chriadam, line 7: typo

lib/webengine.h
```diff
@@ -16,8 +16,19 @@
...
 #include <qmozcontext.h>
 
+/*!
+  \module SailfishWebView
+*/
+
+
 namespace SailfishOS {
 
+/*!
```
chriadam, line 26: Qt style prefers docs in .cpp rather than .h to avoid cluttering .h

doc/index.qdoc
```diff
@@ -0,0 +1,79 @@
...
+CONFIG += link_pkgconfig
+PKGCONFIG += qt5embedwidget
+\endcode
+
+\section2 C++ API
+
+The C++ API consists of the following classes:
+
+\list
+\li the \l{SailfishOS}{WebEngine} class
```
chriadam, line 33: I think these two links might be incorrectly formatted, although I might be wrong

doc/sailfish-webview.cpp
llewelld: Thanks; fixed, plus a couple of others as well.

lib/webengine.h
llewelld: Thanks, I've shifted this to the cpp. As an aside, I've always found it a bit strange that the docs are in the implementation rather than the header file.

doc/index.qdoc
llewelld: I've amended this.

doc/index.qdoc
```diff
@@ -0,0 +1,79 @@
...
+\endcode
+
+\section2 C++ API
+
+The C++ API consists of the following classes:
+
+\list
+\li the \l{SailfishOS::WebEngine}{WebEngine} class
+\li the \l{SailfishOS::WebEngineSettings}{WebEngineSettings} class
+\li the \l{SailfishOS::WebEngineUtils::DownloadHelper}{DownloadHelper} class
```
rainemak, line 35: Let's mark this one as \internal on the docs. It shouldn't really be exposed from the API.

doc/index.qdoc
```diff
@@ -0,0 +1,79 @@
...
+import Sailfish.WebView 1.0
+
+ApplicationWindow {
+    id: root
+    initialPage: Component {
+        Page {
+            id: page
+            WebView {
+                anchors.fill: parent
+                active: true
```
rainemak, line 69: Wondering should we revisit default value state for the WebView i.e. active: true by default.

doc/sailfish-webview.qdocconf
```diff
@@ -0,0 +1,56 @@
+project         = Sailfish OS WebView Component
+description     = Sailfish OS Webview Component Reference Documentation
```
rainemak, line 2: Sailfish WebView Component

rainemak: Similar to Sailfish Silica and Sailfish Pickers for instance.

doc/sailfish-webview.qdocconf
```diff
@@ -0,0 +1,56 @@
...
+qhp.projects = SailfishWebView
+
+qhp.SailfishWebView.file = sailfish-webview.qhp
+qhp.SailfishWebView.namespace = org.sailfishos.wevview.1.0.0
+qhp.SailfishWebView.virtualFolder = sailfishos-webview
+qhp.SailfishWebView.indexTitle = Sailfish OS WebView Component
+qhp.SailfishWebVoew.indexRoot =
+
+qhp.SailfishWebView.filterAttributes = sailfishos-webview 1.0.0
+qhp.SailfishWebView.customFilters.SailfishWebView.name = Sailfish OS WebView 1.0.0
```
rainemak, line 28: Extra OS as well

lib/webengine.cpp
```diff
@@ -16,7 +16,27 @@
...
-void SailfishOS::WebEngine::initialize(const QString &profilePath)
+/*!
+  \class SailfishOS::WebEngine
+  \brief Provides access to the Web engine context.
+  \inmodule SailfishWebView
+
+  Singleton class which provides access to the Web engine context.
+*/
+
+namespace SailfishOS {
```
rainemak, line 27: No no... Sailfish

rainemak: Shit why on earth this was like this at the first place.

doc/index.qdoc
llewelld: Removed from the list and marked as internal.

doc/index.qdoc
llewelld: That would seem to make sense.

doc/sailfish-webview.qdocconf
llewelld: Thanks; I've updated all instances to reflect this.

doc/sailfish-webview.qdocconf
llewelld: Thanks; fixed.

lib/webengine.cpp
llewelld: I can go through and change all instances, but it will affect the library C++ API, which I know is in use by external projects.

rainemak: Let's keep this in mind and consider a bit later. I'm quite confident that we just need to break this. Good thing is that it's simple to fix if it's used.

doc/index.qdoc
```diff
@@ -0,0 +1,78 @@
+/****************************************************************************************
+**
+** Copyright (C) 2021 Open Mobile Platform LLC
+** All rights reserved.
+**
+****************************************************************************************/
+
+/*!
+\page index.html
+\title Sailfish WebView Component
```
rainemak, line 10: I'd actually drop that "Component" as well.

doc/index.qdoc
rainemak: Please file a task under JB#53778 and let's fix the default value for sailfish-components-webview side.

rpm/sailfish-components-webview.spec
```diff
@@ -66,21 +66,30 @@ Requires:  %{name} = %{version}
...
-Summary:    Sailfish OS WebEngine development files
+Summary:    Sailfish WebEngine development files
 Requires:   %{name} = %{version}-%{release}
 
 %description devel
 Development package which provides libsailfishwebengine
 
 %package tests
-Summary:    Sailfish OS WebView / WebEngine tests
+Summary:    Sailfish WebView / WebEngine tests
```
rainemak, line 76: If you want do some additional cleanup, removing "Group:" from the spec would be a nice plus. Not mandatory though.

See:
https://sailfishos.org/wiki/Software_Packaging

doc/index.qdoc
llewelld: Filed as JB#54001.

doc/index.qdoc
llewelld: Have done so.

rpm/sailfish-components-webview.spec
llewelld: Have also done so.

rainemak (approved):

doc/index.qdoc
rainemak: Thanks

doc/index.qdoc
```diff
@@ -0,0 +1,78 @@
+/****************************************************************************************
+**
+** Copyright (C) 2021 Open Mobile Platform LLC
```
atatarov, line 3:
```suggestion
** Copyright (c) 2021 Open Mobile Platform LLC
```
And same for others, @rainemak what do you think?

llewelld: Thanks, that's a good point. I've fixed this (at least for the files this PR touches).

atatarov (approved):

PR #105, opened 2021-04-29
[sailfish-webview] Multiple alert fix. Fixes JB#51913

mkenttala:


import/popups/PopupOpener.qml
```diff
@@ -78,19 +80,13 @@ Timer {
         var winId = data.winId
         switch (topic) {
         case "embed:alert": {
```
rainemak, line 82: Can this get called more than 2 times? Should there be a queue of data objects received from the embedlite?

import/popups/PopupOpener.qml
```diff
@@ -189,6 +185,31 @@ Timer {
...
+                    "winId": data.winId,
+                    "checkvalue": dialog.checkboxValue
+                })
+            })
+        })
+    }
+
+    // Handle pagestack busy change
+    function busyChanged() {
+        if (!pageStack.busy) {
```
rainemak, line 207:
```suggestion
        if (!pageStack.busy && root.alertData) {
```

import/popups/PopupOpener.qml
```diff
@@ -42,6 +42,8 @@ Timer {
         anchor: Notice.Center
     }
 
+    property var alertData
```
rainemak, line 45: Worth creating new tasks to fix confirm and prompt as well as those should be similarly broken. Should checked at least.

import/popups/PopupOpener.qml
mkenttala: This won't be called simultaneously so no queue needed. Fix is only needed to wait for the previous dialog/page to fully close.

import/popups/PopupOpener.qml
mkenttala: Will check

import/popups/PopupOpener.qml
```diff
@@ -189,6 +154,127 @@ Timer {
...
+                    "accepted": false,
+                    "checkvalue": dialog.checkboxValue
+                })
+            })
+        })
+    }
+
+    // Open login dialog
+    function login(data) {
+        var obj = pageStack.animatorPush(Qt.resolvedUrl("PasswordManagerDialog.qml"), {
```
rainemak, line 228:
```suggestion
        pageStack.animatorPush(Qt.resolvedUrl("PasswordManagerDialog.qml"), {
```

rainemak: Please add also testcase for https://github.com/sailfishos/sailfish-browser/blob/next/tests/manual/testwebprompts.html. Simple testcase function that requests few alerts, prompts, and confirms.

rainemak (approved):
Reviewing and testing. Please add manual testcase that runs alerts, confirms, and prompts in row (sailfish-browser).

rainemak:
@atatarov @okodron review as well

PR #108, opened 2021-05-17
Add WebEngine documentation

llewelld:
Applied on top of some of the other documentation commits. I'll rebase it all once they're merged.


doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,159 @@
...
+          }
+      }
+  }
+  \endcode
+
+  \sa WebVew
+*/
+
+/*!
+  \qmlmethod bool WebEngine::isInitialized()
```
rainemak, line 55: We should check if this is really in use... maybe turn into just public C++ method + qml property:

```
Q_PROPERTY(bool initialized READ isInitialized NOTIFY initializedChanged FINAL)
```

Seeing similar glitch in the QMozEngineSettings

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,159 @@
...
+  \brief TBC
+*/
+
+/*!
+  \qmlmethod bool WebEngine::notifyObservers(topic, value)
+  \brief TBC
+*/
+
+/*!
+  \qmlmethod bool WebEngine::runEmbedding(aDelay)
```
rainemak, line 89: This is is the most crucial methods in order to start the engine.

rainemak (commented):
Early review. I need to refresh my memory regarding C++ docs but we should expose both C++ and QML docs.

doc/sailfish-webview-webengine.qdoc
llewelld: Actually, I must admit to being a bit confused about `runEmbedding()`. It's called in `SailfishOS::WebEngine::initialize()`, which is itself called in both `SailfishOSWebViewPlugin::initializeEngine()` and `SailfishOSWebEnginePlugin::initializeEngine()`, which will be run automatically in QML. So there should be no need for a developer to call it directly via QML I think.

So I think it should only appear in the C++ docs, for developers rolling their own initialisation. Does that seem fair?

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,159 @@
...
+  \qmlmethod bool WebEngine::isAccelerated()
+  \brief Returns true if the view is using graphics acceleration, false otherwise.
+
+  Allows the status of graphics acceleration to be queried.
+
+  Graphics acceleration may be Open GL or GLES, for example.
+*/
+
+/*!
+  \qmlmethod bool WebEngine::setIsAccelerated(aIsAccelerated)
```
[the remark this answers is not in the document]
llewelld: Thanks; I'll remove it from the docs.

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,159 @@
...
+  \sa WebVew
+*/
+
+/*!
+  \qmlmethod bool WebEngine::isInitialized()
+  \brief Returns true if the component has been initialized, false otherwise.
+*/
+
+/*!
+  \qmlmethod bool WebEngine::isAccelerated()
```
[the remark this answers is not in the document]
llewelld: Okay, I'll drop `isAccelerated()` also.

doc/sailfish-webview-webengine.qdoc
llewelld: I can turn this into a property and document it as such as I go.

llewelld:
> Early review. I need to refresh my memory regarding C++ docs but we should expose both C++ and QML docs.

Thanks for the useful comments. I'll make sure I expand on the C++ docs as well as the QML ones.

llewelld:
Now rebased so there's only one commit.

lib/webengine.cpp
```diff
@@ -17,24 +17,40 @@
...
+
+    1. Set up the environemnt, including setting \c{GRE_HOME} to the location
+       of the executing binary.
+    2. set up the profile path by calling \l setProfile
+    3. Add any needed component manifests by calling \l addComponentManifest (one
+       call for each manifest).
+
+    After which \l runEmbedding can be called to set the WebEngine event loop
+    running.
+
```
llewelld, line 52: While writing this part I realised that it's not currently really possible to override the default initialisation process after all, because there's no way to set the `isInitialized` value [here](https://github.com/sailfishos/sailfish-components-webview/blob/ebdacfbfa735d8b81233ce46d3a3cd7b33210b95/lib/webengine.cpp#L42), and so the built-initialisation will always happen as well.

Possible solutions to this:
1. Add a `WebEngine::overrideInitialization() { isInitialized = true; }` method that a developer can use to specify they don't want the standard initialisation to run.
2. Check whether `setProfile()` has been already set up and skip the initialisation if it has.
3. Make `WebEngine::initialize()` virtual so that devs can override it.
4. Since `Webengine` is really just an initialisation shim around `QMozContext`, document it in terms of reimplementing `WebEngine` as some other class that inherits `QMozContext`.
5. Don't support overriding the initialization process and don't include this documentation about it (probably in this case `setProfile()` and `runEmbedding()` needn't be included in the docs either??).

Any thoughts?

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,246 @@
...
+    \class QMozContext
+    \brief Web engine context implementation
+    \inmodule SailfishWebView
+
+    \c QMozContext is inherited by \l WebEngine to provide methods for
+    interacting with the underlying EmbedLiteApp and gecko engine.
+
+    Developers should access the methods through \l WebEngine rather than
+    instantiating a \c QMozContext explicitly.
+*/
```
llewelld, line 22: Should these methods be described as part of `QMozContext` and inherited by `WebEngine`, or should we pretend they're just part of `WebEngine`? (I'm not actually sure if we can achieve the latter nicely with qdoc though).

rainemak: I'd document all the relevant API bits here.

lib/webengine.cpp
rainemak: WebEngine::initialized is currently called by QML plugin so for most QML app developers WebEngine is not needed. However, if you need to do custom steps I think documenting WebEngine::initialize would be good.

What if we'd alter initialize method a bit.
```
static void initialize(const QString &profilePath, bool runEmbedding = true);
```
If you need to delay calling the runEmbedding for a reason or another, you could now do it.

You really need those env variables, default components in order the WebView to function correctly. However, one may want to add more components. That said, I still think that more common case is that one needs some extra handling for the loaded content and then a frame script is enough.

Btw, I recently added example teardown of View + Engine to the QtMozEmbed test cases.

From application sandboxing point of view, I think it should be said that qstandard application paths are safe for profile paths as those include both application and organization name.  (FYI for @Tomin1 @okodron ).

doc/sailfish-webview-qmozcontext.cpp
llewelld: I just want to confirm in case I'm misunderstanding: document them here in a dummy `QMozContext` that `WebEngine` inherits?

lib/webengine.cpp
llewelld: Thanks for the suggestion and helpful comments. I've made some changes, including adding the extra parameter and documenting it.

One thing that occurs is that in practice, because the call to `runEmbedding` in `initialize` is wrapped in a timer, in practice these are equivalent:

```
        webEngine::initialize(profilePath, false);
        webEngine->addComponentManifest(QLatin1String("/path/file.manifest"));
        QTimer::singleShot(0, webEngine, SLOT(runEmbedding()));
```
and
```
        webEngine::initialize(profilePath);
        webEngine->addComponentManifest(QLatin1String("/path/file.manifest"));
```
So we could maybe even get away without the extra parameter in this case. Nevertheless my personal feeling is that it's better to include it anyway, just in case there's a circumstance in which it's needed.

doc/sailfish-webview-webengine.qdoc
llewelld: Documented as a property, but I didn't change the code yet (I realise now it's actually in [qtmozembed](https://git.sailfishos.org/mer-core/qtmozembed/blob/0d87492873178eea54af014087499ddb5729a614/src/qmozcontext.h#L36) so I'll create a separate PR).

llewelld: Here's a PR to change `initialized` and `pixelRatio` into properties:

https://git.sailfishos.org/mer-core/qtmozembed/merge_requests/95

llewelld:
Updated to reflect `initialised` and `pixelRatio` being properties.

llewelld:
I've removed `QMozContext::pixelRatio` from the documentation to align with the changes in https://git.sailfishos.org/mer-core/qtmozembed/merge_requests/95.

doc/sailfish-webview-qmozcontext.cpp
rainemak: Yes, and methods/properites that I see relevant are:
```
properties:
initialized

methods (C++ only)
void addObservers(const std::vector<std::string> &aObserversList);
static QMozContext *instance();

slots
void addObserver(const QString &aTopic);
void removeObserver(const QString &aTopic);

void notifyObservers(const QString &topic, const QString &value);
void notifyObservers(const QString &topic, const QVariant &value);

void addComponentManifest(const QString &manifestPath);

void runEmbedding(int aDelay = -1);
void stopEmbedding();
void setProfile(const QString &);

signals:
void contextDestroyed();
void lastViewDestroyed();
void lastWindowDestroyed();
void recvObserve(const QString message, const QVariant data);
``` 

WebEngine does call runEmbedding from its initialization but custom initialization process would require setProfile, addComponentManifest, runEmbedding, and stopEmbedding. 

I.e. would leave out registerWindow, isAccelerated, GetApp, PostUITask, PostCompositorTask, CancelTask, setIsAccelerated, setViewCreator, createView, notifyFirstUIInitialized at least

llewelld: Thanks for the clarifications, that all makes sense. I've updated the PR so that the current roster looks like this. There are minor differences from the list you provided, but hopefully the differences are for valid reasons (e.g. `isInitialized` seems to be a consequence of having the `initialized` property). 
<h2 id="properties">Properties</h2>
<ul>
<li class="fn"><b>initialized</b> : const bool</li>
</ul>
<a name="public-functions"></a>
<h2 id="public-functions">Public Functions</h2>
<div class="table">
<br/>void <b>addObservers</b>(const std::vector&lt;std::string&gt; &amp;aObserversList)
<br/>bool <b>isInitialized</b>() const
<br/>void <b>removeObservers</b>(const std::vector&lt;std::string&gt; &amp;aObserversList)

</div>
<a name="public-slots"></a>
<h2 id="public-slots">Public Slots</h2>
<div class="table">
<br/>void <b>addComponentManifest</b>(const QString &amp;manifestPath)
<br/>void <b>addObserver</b>(const QString &amp;aTopic)
<br/>void <b>notifyObservers</b>(const QString &amp;topic, const QString &amp;value)
<br/>void <b>notifyObservers</b>(const QString &amp;topic, const QVariant &amp;value)
<br/>void <b>removeObserver</b>(const QString aTopic)
<br/>void <b>runEmbedding</b>(int aDelay = -1)
<br/>void <b>setProfile</b>(const QString &amp;profilePath)
<br/>void <b>stopEmbedding</b>()

</div>
<a name="signals"></a>
<h2 id="signals">Signals</h2>
<div class="table">
<br/>void <b>contextDestroyed</b>()
<br/>void <b>initialized</b>()
<br/>void <b>lastViewDestroyed</b>()
<br/>void <b>lastWindowDestroyed</b>()
<br/>void <b>recvObserve</b>(const QString message, const QVariant data)

</div>

llewelld:
Fixed some typos.

doc/sailfish-webview-qmozcontext.cpp
chriadam: It seems strange to have an initialized() signal as well as an initialized property.  Should the signal be initializedChanged()?

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    \sa addObserver, recvObserve
+*/
+void QMozContext::notifyObservers(const QString &topic, const QVariant &value);
+
+/*!
+    \brief Starts the engine event loop.
+
+    When WebEngine is instantiated as a QML component this is called
+    automatically during initalisation. However, it can be called earlier to
```
chriadam, line 131: initalisation -> initialization (missing i, plus american spelling)

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    set the WebEngine event view running in case the intialization process is
```
chriadam, line 132: view -> loop?

chriadam, line 132: intialization -> initialization

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    be created). By default this is set to
+    \c QStandardPaths::writableLocation(QStandardPaths::CacheLocation), which
+    is usually \c{~/.cache/<organisation name>/<application name>}.
+
+    Generally there is no need to call \c setProfile as the profile will
+    automatically set up on initialization, and can also be passed as
+    parameter to \l{WebEngine::initialize}.
+
+    When WebEngine is instantiated as a QML component this is also called
+    automatically during initalisation. However, it can be called earlier to
```
chriadam, line 178: initalisation -> initialization

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    set a custom profile location in case the intialization process is being
```
chriadam, line 179: it's a difficult one to spell ;-)

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    overridden.
+
+    \note If using a different location care is needed to select a path that's
+    acceptable from an application sandboxing point of view. The qstandard
```
chriadam, line 183: maybe link to them: The \l{https://doc.qt.io/archives/qt-5.6/qstandardpaths.html}{standard application paths} ...

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+/*!
+    \brief This signal is sent when the last registered window has been destroyed.
+
+    By default this is used to delay execution of \l stopEmbedding until after
+    the last window has closed.
+*/
+void QMozContext::lastWindowDestroyed();
+
+/*!
+    \brief Regsiter javascript chrome components to be loaded into the WebEngine.
```
chriadam, line 250: Regsiter -> Register

chriadam: Also, maybe we should use camel-casing for javascript -> JavaScript ?

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    The sailfish WebEngine has three main layers. The top layer is exposed as
```
chriadam, line 252: Sailfish should be capitalised, I think.

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    QML components with a developer-facing interface. The lowest level is the
+    native gecko rendering engine.
+
+    Between the two sits a javascript chrome layer that can communicate with
```
chriadam, line 256: What is meant by "chrome layer"?  I'd prefer something more expressive, maybe "a content and event processing layer defined in JavaScript" but I'm not sure.

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+    either of the other layers using an observer/notification pattern. These
+    javascript components are considered privileged code and run with the
+    \l {https://developer.mozilla.org/en-US/docs/Mozilla/Gecko/Script_security#security_principals}{system principal}.
```
chriadam, line 259: This worries me slightly - is there a specific permission that the app has to request, in order to use this?  I don't like the idea of an application showing me a web view, which can read credentials/content from sites accessed via the webview... I mean, I guess a normal webview is loaded in-process anyway, so it's possible to read the contents of memory regardless, but still ...

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+              break;
+            case "exampleTopic":
+              Logger.debug("ExampleComponent: exapleTopic data: " + aData);
+              break;
+            }
+          },
+        };
+
+        this.NSGetFactory = XPCOMUtils.generateNSGetFactory([ExampleComponent]);
+    \endcode
```
chriadam, line 340: nice code example.  I wonder if some description of possible use-cases could be included below, as text.  I mean, why would app developers want to do this, in general?

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,367 @@
...
+*/
+void QMozContext::addComponentManifest(const QString &manifestPath);
+
+/*!
+    \property QMozContext::initialized
+    \brief Whether the context has been initialized
+
+    This property returns true when the context has been initialized and false
```
chriadam, line 348: I think this sentence is unnecessary, personally.

doc/sailfish-webview-qmozcontext.h
```diff
@@ -0,0 +1,41 @@
+/****************************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC
+** All rights reserved.
+**
+****************************************************************************************/
+
+#include <QObject>
+
+// Used only for documentation purposes
```
chriadam, line 10: Instead of "Used only for documentation purposes" maybe "Dummy declaration for documentation purposes only".

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,326 @@
...
+** Copyright (c) 2021 Open Mobile Platform LLC
+**
+****************************************************************************/
+
+/*!
+    \qmltype WebEngine
+    \inherits Item
+    \inqmlmodule Sailfish.WebEngine
+    \ingroup sailfishos-webview
+    \brief Provides access to the Web engine context.
```
chriadam, line 12: maybe remove capitalisation from Web

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,326 @@
...
+    \sa WebView
+*/
+
+/*!
+    \readonly
+    \qmlproperty bool WebEngine::initialized
+    \brief Returns true if the component has been initialized, false otherwise.
+*/
+
+/*!
```
chriadam, line 60: Hrm, the explicit duplication of documentation for C++ and QML sides is ... unfortunate.  However, I don't have any suggestion for how to resolve it..

lib/webengine.cpp
```diff
@@ -17,26 +17,60 @@
...
+    \code
+        SailfishOS::WebEngine::instance() *webEngine = SailfishOS::WebEngine::instance();
+        QString profilePath = QStandardPaths::writableLocation(QStandardPaths::CacheLocation);
+        webEngine->initialize(profilePath, false);
+        webEngine->addComponentManifest(QLatin1String("/path/file.manifest"));
+
+        [...]
+
+        QTimer::singleShot(0, webEngine, SLOT(runEmbedding()));
```
chriadam, line 57: ... or `QMetaObject::invokeMethod(webEngine, "runEmbedding", Qt::QueuedConnection);`

lib/webengine.cpp
```diff
@@ -67,39 +101,44 @@ void WebEngine::initialize(const QString &profilePath)
...
     isInitialized = true;
 }
 
 /*!
-  \brief Returns the instance of the singleton WebEngine class.
+    \brief Returns the instance of the singleton WebEngine class.
 
-  Returns the instance of the singleton WebEngine class.
-  The returned instance may not be intialised.
+    The returned instance may not be intialised.
```
chriadam, line 114: intialised -> initialized

lib/webengine.cpp
```diff
@@ -67,39 +101,44 @@ void WebEngine::initialize(const QString &profilePath)
...
     : QMozContext(parent)
 {
 }
 
 /*!
-  \brief Destroys the WebEngine instance.
+    \internal
+    \brief Destroys the WebEngine instance.
+
+    WebEngine is a singleton, so should never be deleted.
```
chriadam, line 141: .. by the client.  It will be automatically deleted by the QML engine at application shutdown.

doc/sailfish-webview-qmozcontext.cpp
llewelld: That's a good question; we had some discussion about it [here](https://git.sailfishos.org/mer-core/qtmozembed/merge_requests/95#note_89619). Use of `initialized()` for the signal predates the change to a property, so I kept it like that to avoid complications. It could be revisited. For QML Raine discovered that `onInitializedChanged` still works, but for the C++ side I guess the real signal name is needed.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks; fixed.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Yes, correct, thank you. Now fixed.

llewelld: Fixed.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Fixed.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks for highlighting them all. Not sure why my fingers have so much trouble with it.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks, good idea. Added in.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks. I've updated all of the JavaScripts too.

doc/sailfish-webview-qmozcontext.cpp
llewelld: I guess I was thinking this is where all the XUL stuff used to happen. Rather than wrestle with what it really is, I've just changed it to "JavaScript layer".

doc/sailfish-webview-qmozcontext.cpp
llewelld: There's no permission. What you say is accurate and I hear your concern, but realistically I wonder if there's any way to avoid this. Even if this capability weren't available the app could display a fake local copy of a website, ask the user to enter their credentials, capture the credentials, then forward the user to the real website, all without the user knowing. As there's no "trusted TLS padlock" in the webview, the user wouldn't be able to tell the difference. I presume this is why Android doesn't allow OAuth to be performed inside apps any more (to discourage users from entering their credentials into anything except the system browser).

doc/sailfish-webview-qmozcontext.cpp
llewelld: I've added in some text to try to explain. Do you think I should be even more explicit and give a tangible example (e.g. using it to upgrade all connections to TLS, say)?

doc/sailfish-webview-qmozcontext.cpp
llewelld: Yes, it is a bit. I've removed it.

doc/sailfish-webview-qmozcontext.h
llewelld: Thanks; I've changed it.

doc/sailfish-webview-webengine.qdoc
llewelld: Removed (even though it's a bit of a struggle accepting that web's not considered a proper noun any more; old habits).

doc/sailfish-webview-webengine.qdoc
llewelld: I'm happy to act on any suggestions, but I don't have a solution either I'm afraid.

lib/webengine.cpp
llewelld: I've taken this directly from [the initialization code](https://github.com/sailfishos/sailfish-components-webview/blob/236086301dd9be6b4089b78278114c65f6069549/lib/webengine.cpp#L70), which I'd prefer to duplicate as it's instructive. Is there a good reason to prefer `invokeMethod` here (in which case, should we change both)?

lib/webengine.cpp
llewelld: Thanks; I've changed this.

doc/sailfish-webview-qmozcontext.cpp
rainemak: Originally I was thinking also ```initializedChanged``` but as generated slot (onInitializedChanged) is still there on QML due to signal name I think this is actually better. 

As this is singleton and should not be initialized explicitly following example is better than only having initializedChanged signal.

```
    Connections {
        target: WebEngine

        onInitialized: console.log("======== Do something")
```
At the same time, we have quite a few places where we are using ```onInitialized:``` and I do think that having WebEngine, WebView, WebEngineSettings in sync is the key factor and eases the API usage.

When going forward, at some point we could consider adding ```WebComponent``` component and attached signals for  initialized and destroyed. Something similar to ```Component.onCompleted``` => ```WebComponent.onInitialized``` => Out of scope for now I think.

doc/sailfish-webview-qmozcontext.h
```diff
@@ -0,0 +1,41 @@
...
+**
+** Copyright (c) 2021 Open Mobile Platform LLC
+** All rights reserved.
+**
+****************************************************************************************/
+
+#include <QObject>
+
+// Dummy declaration for documentation purposes only
+class QMozContext {
```
rainemak, line 11: I was thinking that we would mask QMozContext away from the docs altogether. I was thinking to leave open an option to merge qtmozembed to webview in long run. But well, that's very unlikely to happen => this works as well. If we need to change later, then we change later.

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    \sa {SailfishOS::WebEngine::initialize}{WebEngine::initialize}, runEmbedding, setProfile
+*/
+void QMozContext::stopEmbedding();
+
+/*!
+    \internal
+    \brief Sends the "final-ui-startup" notification to observers
+
+    For internal use? I don't actually see it used anywhere.
```
rainemak, line 160: I vaguely remember that this will be called by some xre function inside gecko upon first startup. But definitely this should be kept internal.

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,380 @@
...
+    The \a message parameter will be the topic in question, the \a data parameter
+    will contain data sent by the sender constructed from a JSON object. The
+    exact object structure is at the sender's discretion.
+
+    \sa addObserver, addObservers, removeObserver, removeObservers, notifyObservers
+*/
+void QMozContext::recvObserve(const QString message, const QVariant data);
+
+/*!
+    \brief This signal is emitted when the gecko EmbedLiteApp has been destroyed.
```
rainemak, line 230: I'd not add references to the EmbedLiteApp and friends as those are implementation details.
```
\brief This signal is emitted after embedding has been \l{stopEmbedding}{stopped} and context is ready to be deleted.  
```

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+// Dummy immplementation for documentation purposes only
+
+/*!
+    \class QMozContext
+    \brief Web engine context implementation.
+    \inmodule SailfishWebView
+
+    \c QMozContext is inherited by \l WebEngine to provide methods for
+    interacting with the underlying EmbedLiteApp and gecko engine.
```
rainemak, line 19:
```suggestion
    interacting with the underlying Gecko browser engine.
```

doc/sailfish-webview-qmozcontext.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+    Developers should access the methods through \l WebEngine rather than
+    instantiating \c QMozContext directly.
+*/
+
+/*!
+    \brief Registers the WebEngine for receiving notifications on a topic.
+
+    The gecko and EmbedLite components use notifications to transfer data and
```
rainemak, line 28: EmbedLite components are fine to be mentioned as those are part of low-level integration mechanisms.

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,327 @@
...
+
+    Additional data can be sent with the notification using the \a value
+    parameter in JSON format.
+
+    \sa addObserver, recvObserve
+*/
+
+/*!
+    \internal
+    \qmlmethod WebEngine::runEmbedding(aDelay)
```
rainemak, line 213: This is on of those methods that I'd have liked to see in WebEngine C++ doc side which I referred here https://github.com/sailfishos/sailfish-components-webview/pull/108/files#r639432191. Surely, not good for QML side.

doc/sailfish-webview-qmozcontext.h
llewelld: I can move all of the relevant `QMozContext` documentation into the `WebView` C++ docs (using the `#ifdef Q_QDOC`' approach Chris suggested). But I perhaps misunderstood an earlier comment as I thought you preferred them in separate docs. Should I change it?

doc/sailfish-webview-qmozcontext.cpp
llewelld: I've removed the additional comment and left this with the brief text and marked as `\internal`.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Okay, yes fair enough. I've removed the reference and used the text you suggested.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks; fixed.

doc/sailfish-webview-qmozcontext.cpp
llewelld: Thanks: left as-is.

doc/sailfish-webview-qmozcontext.h
rainemak: Let's put this way, it's safer for the future if all is documented under the WebEngine. Like mentioned elsewhere, I do have my doubts that it happens any time soon. Hence, this is fine as well.

My original point was to document all under `WebEngine` itself (primary) and if Q_QDOC would be a tool if needed (secondary). No worries @llewelld. This already super good.

In order to sort this out neat way, I think we could add to the QMozEmbed docs a disclaimer that this may more to WebEngine itself in future. This way, it'd give us more breathing space in future. Further, as this should not really be used too heavily by 3rd party dev, I hope & think it's not a real issue even either, this should be fine like this as well.

lib/webengine.cpp
```diff
@@ -17,26 +17,61 @@
...
 /*!
-  \brief Initialises the WebEngine class.
+    \brief Initialises the WebEngine class.
 
-  Initalises the WebEngine class. The \a profilePath sets the root folder for
-  the mozilla gecko profile. Once set the \a profilePath can't be changed.
+    Initalizes the WebEngine class. The \a profilePath sets the root folder for
+    the mozilla gecko profile. Once set the \a profilePath can't be changed.
+    Set \a runEmbedding to \c false (defaults to true) in order to delay
+    execution of the EmbedLite event loop. In this case, \l runEmbedding must
```
rainemak, line 36: Something like:
```
in order to handle steps that are needed before \l{engine startup}{runEmbedding}.
``` 
Point is that that let's try to avoid "EmbedLite event loop".  I think after the change "In this case, \ runEmbedding..." can be removed altogether.

doc/sailfish-webview-qmozcontext.h
llewelld: Thanks for the explanation. I've added in a warning that developers should access the methods through the WebEngine interface as they may be moved there in future.

lib/webengine.cpp
llewelld: Thanks; I've updated this with your suggestion; it looks clear to me.

llewelld:
Rebased and addressed conflicts.

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,328 @@
...
+        import Sailfish.Browser 1.0
+        import Sailfish.WebEngine 1.0
+
+        ApplicationWindow {
+            id: root
+            initialPage: Component {
+                WebViewPage {
+                    WebView {
+                        anchors.fill: parent
+                        active: true
```
rainemak, line 97: This is can be removed.

llewelld: Thanks! I've plucked it from here, and a couple of other places too.

llewelld:
I've made all of the QMozContext docs part of WebEngine instead. It's in a separate commit, so I can revert it if it's a retrograde step. It certainly makes things clearer and simpler.

doc/sailfish-webview-webengine.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    \sa setViewCreator
+*/
+void SailfishOS::WebEngine::createView();
+
+/*!
+    \brief This signal is emitted when a notification on an observed topic is received.
+
+    After an app has registered an interest in a particular topic by calling
+    \l addObserver or \l addObservers, it can then conect to this signal to be
```
rainemak, line 210:
```suggestion
    \l addObserver or \l addObservers, it can then connect to this signal to be
```

doc/sailfish-webview-webengine.cpp
```diff
@@ -0,0 +1,362 @@
...
+*/
+void SailfishOS::WebEngine::contextDestroyed();
+
+/*!
+    \brief This signal is emitted when when the last view has been destroyed.
+*/
+void SailfishOS::WebEngine::lastViewDestroyed();
+
+/*!
+    \brief This signal is sent when the last registered window has been destroyed.
```
rainemak, line 237:
```suggestion
    \brief This signal is emitted when the last registered window has been destroyed.
```

doc/sailfish-webview-webengine.cpp
```diff
@@ -0,0 +1,362 @@
...
+        #include <sailfishapp.h>
+        #include <libsailfishwebengine/webengine.h>
+
+        int main2(int argc, char *argv[])
+        {
+            QGuiApplication *app = SailfishApp::application(argc, argv);
+            QQuickView *view = SailfishApp::createView();
+            view->setSource(SailfishApp::pathTo("application.qml"));
+
+            SailfishOS::WebEngine::instance() *webEngine = SailfishOS::WebEngine::instance();
```
rainemak, line 280:
```suggestion
            SailfishOS::WebEngine *webEngine = SailfishOS::WebEngine::instance();
```

doc/sailfish-webview-webengine.qdoc
```diff
@@ -0,0 +1,325 @@
...
+    \qmlsignal WebEngine::lastViewDestroyed()
+    \brief This signal is emitted when when the last view has been destroyed.
+
+    In practice it doesn't seem to be emitted anywhere.
+*/
+
+/*!
+    \internal
+    \qmlsignal WebEngine::lastWindowDestroyed()
+    \brief This signal is sent when the last registered window has been destroyed.
```
rainemak, line 319:
```suggestion
    \brief This signal is emitted when the last registered window has been destroyed.
```

lib/webengine.cpp
```diff
@@ -17,26 +17,60 @@
...
+    This method will be called automatically during QML initialisation of the
+    WebView. However under some circumstances it may be useful to perform a
+    custom initalization. In this case \c initialize can be called manually as
+    long as this id done before the QML engine initialization.
+
+    One reason for custom initialization would be to proivde additional
+    component manifests.
+
+    \code
+        SailfishOS::WebEngine::instance() *webEngine = SailfishOS::WebEngine::instance();
```
rainemak, line 49:
```suggestion
        SailfishOS::WebEngine *webEngine = SailfishOS::WebEngine::instance();
```

llewelld:
Fixed up with the suggestions (thanks Raine), squashed and rebased.

doc/sailfish-webview-webengine.cpp
```diff
@@ -0,0 +1,362 @@
...
+** All rights reserved.
+**
+****************************************************************************************/
+
+#include "sailfish-webview-webengine.h"
+
+// This file is used only for documentation purposes
+
+// These items are all part of QMozContext, but inherited by WebEngine, so we
+// take some notational liberty and present them as part of the the subclass.
```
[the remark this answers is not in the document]
llewelld: Nicely spotted. Thank you.

doc/sailfish-webview-webengine.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+    When WebEngine is instantiated as a QML component this is called
+    automatically during initialization. However, it can be called earlier to
+    set the WebEngine event loop running in case the initialization process is
+    being overridden.
+
+    See \l {SailfishOS::WebEngine::initialize}{WebEngine::initialize} for more info.
+
+    The \a aDelay allows for an asynchronous startup. Calling without the
+    \a aDelay parameter specified will cause execution to happen on the nested
```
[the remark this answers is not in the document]
llewelld: Yes, that's a good point. I added some text to clarify.

rainemak (approved):
Definitely good enough quality docs. I don't have anything else to add. How about you @okodron ?

okodron (approved):

PR #104, opened 2021-04-29
Allow clients to customize the appearance and behavior of popups

chriadam:

PR #109, opened 2021-05-19
[webview] Make it possible to use Silica Page as parent. Fixes JB#54299 OMP#JOLLA-145

rainemak:


rainemak:
Sample app created in context of the https://github.com/sailfishos/sailfish-components-webview/pull/104 uses Silica Page as parent of the WebView and that breaks more or less everything. I think we can have Silica Page as a fallback but we could document things so that WebViewPage should be used.

That said, I do not remember why we have this WebViewPage at the first place - just to have orthogonal API (WebViewPage, WebViewFlickable, and WebView)? Maybe that's a strong enough reason. It's simple to have there now and we can internally add new bits to the WebViewPage as per need.

chriadam:
I don't recall why we had the WebViewPage either... This change seems fine to me, in any case.  I agree, let's not remove WebViewPage at this point, in case it turns out we do need to add something there, in future.

chriadam (approved):
LGTM

okodron (approved):

PR #114, opened 2021-06-09
Update wording on password save pop-up

llewelld:
The original wording was a bit tricky to parse with both username and domain in the sentence, so the username has been removed.

Some new text explaining how passwords can be managed has been added.

### DIFF
--- a/import/popups/PasswordManagerDialog.qml
+++ b/import/popups/PasswordManagerDialog.qml
@@ -34,60 +34,75 @@ Dialog {
                 _glassOnly: true
             }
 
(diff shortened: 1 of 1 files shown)


llewelld:
This is to update the wording on the save login pop-up to:

"Would you like to save Login and password for <domain>?"
"Logins and passwords can be managed from the Settings page"

Screenshot:

login-popup

sledges (approved):

okodron (approved):

rainemak (approved):

PR #115, opened 2021-06-17
[sailfish-components-webview] Implement BlockedPopupInterface. Contributes to JB#54486 OMP#JOLLA-173

chriadam:


chriadam:
Related to: https://github.com/sailfishos/sailfish-browser/pull/867

chriadam:
Is someone able to give me step-by-step instructions on how to test / verify this?  I am not certain how to ensure that popups are blocked by default.

atatarov:
> Is someone able to give me step-by-step instructions on how to test / verify this? I am not certain how to ensure that popups are blocked by default.

I use https://webbrowsertools.com/popup-blocker/ 
Method #5 
Open a popup by creating a dynamic iframe and executing window.open(...) on its context with a delay

atatarov:
1 Case
1) Open popup-test
2) Deny 
3) reload page - popup request should appear again
2 Case
1) Open popup-test  
2) Deny with "rememeber" 
3) reload page - popup requet shouldn't appear
4) Go to settings - permissions - Allow popup global permissions
5) check popup-test - request shouldn't appear
6) Block popup global permissions

3 Case
1) open test
2) Allow popup 
3) request shouldn't appear (popup should working after reload)
4 Case
1) open test
2) Allow popup with "remember"
3) Go to settings, change Block/Allow global permission 
4) The popup should work anyway

chriadam:
Thanks very much for the detailed test instructions!
I have now updated the PR with a variety of fixes for bugs I found during testing.
I think this PR is now ready for review.

import/popups/BlockedPopupDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+                cancelText: qsTrId("sailfish_components_webview_popups-he-block_popup")
+            }
+
+            Label {
+                x: Theme.horizontalPageMargin
+                width: parent.width - x * 2
+
+                //: %1 is the site that wants to load a popup
+                //% "The site %1 has requested to open a new tab. Would you like to allow it to open new tabs?"
+                text: qsTrId("sailfish_components_webview_popups-la-popup_request").arg(blockedPopup.host)
```
atatarov, line 46: There is a already have blocked the opening new tab, and it isn't request, may be should change this message?
What do you think?

chriadam: Good point, yes.  Perhaps something like:
`"The site %1 attempted to open a new tab, which has been blocked.  Would you like to allow it to open new tabs?"`
instead?

atatarov: It's ok for me

chriadam: Done, thanks.

atatarov (approved):

import/popups/BlockedPopupDialog.qml
```diff
@@ -0,0 +1,63 @@
...
+                //% "Deny"
+                cancelText: qsTrId("sailfish_components_webview_popups-he-block_popup")
+            }
+
+            Label {
+                x: Theme.horizontalPageMargin
+                width: parent.width - x * 2
+
+                //: %1 is the site that wants to load a popup
+                //% "The site %1 attempted to open a new tab, which has been blocked. Would you like to allow it to open new tabs?"
```
llewelld, line 45: Currently the `host` value is actually the url of the page being loaded rather than the host. It might be worth updating the wording to "page" instead of "site", or alternatively getting `EmbedLiteChromeManager.js` to pass the host instead.

It might also be worth considering truncating the value to some fixed length (since the page isn't scrollable, a really long domain or url will cause the "remember" button to get lost).

llewelld (commented):
I've made just a few minor suggestions about the documentation, but this otherwise looks good to me. The docs and example are really nice.

When I tested it out I noticed one unexpected behaviour, but I'm not sure whether it's related to this PR, a bug, or intentional.

The following sequence:
1. Open https://browser.sailfishos.org/tests/testpopupwindows.html
2. In the blocker dialog that opens, activate "Remember for this site".
3. Select "Deny".
4. Select the "Open a pop-up window after 2s delay" link.
5. Wait 2 seconds.

On my device the blocker dialog opens again, when I'd expect the popup request to be silently ignored given that I selected the "Remember" option.

doc/sailfish-webview-popups-blockedpopupinterface.qdoc
```diff
@@ -0,0 +1,74 @@
...
+
+  The \c{BlockedPopupInterface} QML type is exposed through the
+  \c{Sailfish.WebView.Popups} QML import namespace.
+
+  Clients who use a \l{WebView} in their application may wish to
+  customize the appearance or behavior of the blocked popup in their
+  application.
+
+  To do so, their implementation must extend the \c{BlockedPopupInterface},
+  wand then they must tell the \l{WebView} to use their implementation via
```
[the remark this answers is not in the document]
chriadam: thanks, fixed

doc/sailfish-webview-popups-blockedpopupinterface.qdoc
```diff
@@ -0,0 +1,74 @@
...
+
+  After the \c{accepted} or \c{rejected} signal is emitted, the value of
+  this property will be used to determine whether to save the decision
+  in the permission manager.
+*/
+
+/*!
+  \qmlsignal BlockedPopupInterface::accepted()
+  \brief This signal is emitted by the popup if the user wishes to allow
+         the site to show the popup.
```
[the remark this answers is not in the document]
chriadam: done

doc/sailfish-webview-popups-blockedpopupinterface.qdoc
```diff
@@ -0,0 +1,74 @@
...
+  After emitting this signal, the popup should perform an appropriate
+  transition to make itself invisible and non-interactable (e.g. for an
+  item-derived popup, an animated opacity transition, and then setting
+  \c{visible} to \c{false} once the animation is complete).
+*/
+
+/*!
+  \qmlsignal BlockedPopupInterface::rejected()
+  \brief This signal is emitted by the popup if the user wishes to disallow
+         the site from showing the popup.
```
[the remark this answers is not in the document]
chriadam: done

examples/custompopups/CustomBlockedPopup.qml
```diff
@@ -0,0 +1,58 @@
...
+    anchors.centerIn: parent
+
+    rememberValue: toggle.checked
+
+    property string acceptText: "Allow"
+    property string cancelText: "Deny"
+
+    Rectangle {
+        anchors.fill: parent
+        color: "chartreuse"
```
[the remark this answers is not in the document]
chriadam: not as good as lightsteelblue but I wanted to use different colours for each different example popup ;-)

import/popups/BlockedPopupDialog.qml
chriadam: changed to "page".  The "host vs page url" discussion is important, but slightly out of scope for this task/PR, I think.

chriadam: Regarding the truncation: what it is a good way to do this, considering that it should scale appropriately for different screen sizes and form factors?

chriadam:
Thanks for noticing that bug.  I had fixed the "accept" branch but not the "deny" branch, I think.  Fixed both now.

chriadam:
> Thanks for noticing that bug. I had fixed the "accept" branch but not the "deny" branch, I think. Fixed both now.

Hrm.  Actually... even with my change, this behaviour still occurs.  The reason is that the popupblocked message is emitted when the popup is blocked, not merely when a prompt is required...

That is, even if a permanent rule exists to say "Deny", we still show the popup.  I will look into fixing this properly, now.

rainemak:
> > Thanks for noticing that bug. I had fixed the "accept" branch but not the "deny" branch, I think. Fixed both now.
> 
> Hrm. Actually... even with my change, this behaviour still occurs. The reason is that the popupblocked message is emitted when the popup is blocked, not merely when a prompt is required...
> 
> That is, even if a permanent rule exists to say "Deny", we still show the popup. I will look into fixing this properly, now.

If not related to this change, please create a new task. As @llewelld has been reviewing this, I'll let him to finish reviewing.

chriadam:
> > > Thanks for noticing that bug. I had fixed the "accept" branch but not the "deny" branch, I think. Fixed both now.
> > 
> > 
> > Hrm. Actually... even with my change, this behaviour still occurs. The reason is that the popupblocked message is emitted when the popup is blocked, not merely when a prompt is required...
> > That is, even if a permanent rule exists to say "Deny", we still show the popup. I will look into fixing this properly, now.
> 
> If not related to this change, please create a new task. As @llewelld has been reviewing this, I'll let him to finish reviewing.

It's unrelated to the task (in terms of simply moving the dialog from sailfish-browser to sailfish-components-webview).
It's related in terms of which code will need to be changed.

Given that it's a separate task, I will create a new bug and we can do the work separately.

llewelld (approved):
Thanks for tweaking the documentation, at all reads quite clearly now. This all LGTM.

Concerning the truncation, I'm not sure what a good way to do this is. In other parts of the underlying gecko code, the host is just [truncated to 150 characters](https://github.com/sailfishos/gecko-dev/blob/66249d15cd1213d67b0a44c9e6c4491260d114b8/toolkit/components/prompts/src/nsPrompter.js#L259):
```
        // Trim obnoxiously long realms.
        if (realm.length > 150) {
            realm = realm.substring(0, 150);
            // Append "..." (or localized equivalent).
            realm += this.ellipsis;
        }
```
This is admittedly nasty, but probably not as nasty as not trimming at all. Alternatively, adding a SilicaFlickable to the dialog might be another way to go. I'm approving, but also happy to review again if you decide to make any further changes.

PR #116, opened 2021-06-24
[webview] Remove spacing between header and list view content in folder picker pages. Contributes to JB#54754 OMP#JOLLA-209

chriadam:


llewelld:
I'm thinking it might make sense to have `headerSpacing` set to 0 by default. In that case this PR and [sailfish-browser#877](https://github.com/sailfishos/sailfish-browser/pull/877/files) would become redundant (not that that's a reason to do it, just a side effect).

chriadam:
This one is no longer required, as the default header spacing has now been set to 0 in the sailfish-components-pickers PR

PR #117, opened 2021-06-24
Support translation keys in dialogs

llewelld:
Updates the default dialogs to use translation keys passed from the gecko engine for localisation of messages presented to the user.

Updates the documentation with info about the translation keys for custom dialog creation.

Adds a customisable Select custom dialog type for selecting multiple values in a dialog with a message.

### DIFF
--- a/doc/sailfish-webview-popups-authpopupinterface.qdoc
+++ b/doc/sailfish-webview-popups-authpopupinterface.qdoc
@@ -26,6 +26,37 @@
 */
 
 /*!
+  \qmlproperty var AuthPopupInterface::messageBundle
+  \brief The gecko localization bundle of the message to display to the user
+
+  A custom implementation can use this information to tailor the message
+  and details supplied to the user.
+
+  The value is either a string or an array. If it's an array then the first
+  value will be the gecko localization key taken from the commonDialogs.properties
+  file. The remaining items in the array are the parameter values to replace %1,
+  %2, etc. in the string.
+
+  If the value is a string then it will be a gecko localization key without
+  parameters.
+
+  You can search the \l{https://pontoon.mozilla.org/en-GB/firefox/}{Firefox
+  Pontoon localization database} for this file, followed by the appropriate key.
+  The key can be one of the following values:
+
+  \list
+    \li EnterLoginForRealm3
+    \li EnterLoginForProxy3
+    \li EnterUserPasswordFor2
+    \li EnterUserPasswordForCrossOrigin2
+    \li EnterPasswordFor
+  \endlist
+
+  Care must be taken as it's possible these values may change with future
+  engine updates.
+*/
+
+/*!
   \qmlproperty string AuthPopupInterface::hostname
   \brief The hostname of the site which is asking for credentials
 
@@ -117,6 +148,35 @@
   \brief The value which should be prefilled into the "remember" toggle switch
 */
 
+
+/*!
+  \qmlproperty var AuthPopupInterface::rememberMessageKey
+  \brief The gecko localization key of the message to display to the user
+         for the "remember" toggle switch
+
+  A custom implementation can use this information to tailor the message
+  shown alongside the "remember" toggle switch.
+
+  The value is either a string or an array. If it's an array then the first
+  value will be the gecko localization key taken from the commonDialogs.properties
+  file. The remaining items in the array are the parameter values to replace %1,
+  %2, etc. in the string.
+
+  If the value is a string then it will be a gecko localization key without
+  parameters.
+
+  You can search the \l{https://pontoon.mozilla.org/en-GB/firefox/}{Firefox
+  Pontoon localization database} for this file, followed by the appropriate key.
+
+  Currently there is only one value the key can take:
+
+  \list
+    \li rememberButton
+  \endlist
+
+  Care must be taken as it's possible these values may change with future
+  engine updates.
+*/
 
 /*!
   \qmlproperty string AuthPopupInterface::usernameValue
--- a/doc/sailfish-webview-popups-authpopupinterface.qdoc
+++ b/doc/sailfish-webview-popups-authpopupinterface.qdoc
@@ -9,6 +9,7 @@
   \inqmlmodule Sailfish.WebView.Popups
   \ingroup sailfishos-webview-popups
   \brief A component which defines the interface for the auth popup
+  \inherits UserPromptInterface
 
   The \c{AuthPopupInterface} QML type is exposed through the
   \c{Sailfish.WebView.Popups} QML import namespace.
@@ -21,6 +22,11 @@
   which itself extends the \l{UserPromptInterface}, and then they must
   tell the \l{WebView} to use their implementation via the \c{popupProvider}
   property.
+
+  The key assigned to \l{Sailfish.WebView.Popups::UserPromptInterface::acceptText}
+  {acceptText} will always be "AcceptLogin" while
+  \l{Sailfish.WebView.Popups::UserPromptInterface::cancelText}{cancelText}
+  will always be assigned the empty string.
 
   \sa UserPromptInterface, custompopups
 */
--- a/doc/sailfish-webview-popups-popupprovider.qdoc
+++ b/doc/sailfish-webview-popups-popupprovider.qdoc
@@ -122,3 +122,11 @@
   The component specified must implement the
   \l{BlockedTabPopupInterface}{blocked tab popup} interface.
 */
+
+/*!
+  \qmlproperty var PopupProvider::selectorPopup
+  \brief The selector popup component descriptor
+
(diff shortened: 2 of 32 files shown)


llewelld:
Requires https://github.com/sailfishos/embedlite-components/pull/5

doc/sailfish-webview-popups-authpopupinterface.qdoc
```diff
@@ -25,6 +25,31 @@
   \sa UserPromptInterface, custompopups
 */
 
+/*!
+  \qmlproperty string AuthPopupInterface::messageKey
```
chriadam, line 29: is "key" the terminology used in gecko/mozilla?  In Qt we use "id" (e.g. translation id) rather than "key" but obviously we should be consistent with gecko terminology in this case.

doc/sailfish-webview-popups-selectpopupinterface.qdoc
```diff
@@ -0,0 +1,67 @@
+/****************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC
+**
+****************************************************************************/
+
+/*!
+  \qmltype SelectPopupInterface
```
chriadam, line 8: maybe SelectorPopupInterface would be a better name?

import/popups/AuthDialog.qml
```diff
@@ -38,6 +40,45 @@ Dialog {
     property alias preventDialogsPrefillValue: auth.preventDialogsPrefillValue
     property alias preventDialogsValue: auth.preventDialogsValue
 
+    // Trim obnoxiously long realms
+    property string trimmedRealm: realm.length <= 150 ? realm : realm.substring(0, 150) + "\u2026"
```
chriadam, line 44: Hmm.  The "obnoxious" length might depend on device type, orientation, etc.  I think using a magic value here might not make the most sense, but instead maybe a constant from dconf would be better?  It's not a huge deal, obviously, but otherwise this is one of those "variation points" which isn't obvious.

import/popups/PasswordManagerDialog.qml
```diff
@@ -16,26 +16,61 @@ Dialog {
...
 
         Item {
             anchors.fill: parent
 
             DialogHeader {
+                id: header
                 //: Accept browser's request to save entered password
                 //% "Save"
-                acceptText: qsTrId("sailfish_components_webview_popups-he-accept_password_mgr_request")
+                acceptText:
```
chriadam, line 66: missing a qsTrId() here I think

import/popups/PopupOpener.qml
```diff
@@ -340,6 +359,33 @@ Timer {
         }
     }
 
+    function selection(data) {
```
chriadam, line 362: maybe selector()

import/popups/PopupOpener.qml
```diff
@@ -366,6 +412,7 @@ Timer {
 
         var obj = null
         if (isDialog) {
+            console.log("openPopup: dialog");
```
chriadam, line 415: debugging logs should be removed I think, a couple more below also

doc/sailfish-webview-popups-authpopupinterface.qdoc
llewelld: At least "key" is the term that's used in the files I've been touching. I've essentially been pushing calls to `getLocalizedString(key, formatArgs)` [like this](https://github.com/sailfishos/gecko-dev/blob/66249d15cd1213d67b0a44c9e6c4491260d114b8/toolkit/components/prompts/src/nsPrompter.js#L115) up into the QML layer.

Admittedly I didn't give it much thought before and I wouldn't mind using `id` instead if this would be clearer in the context?

doc/sailfish-webview-popups-selectpopupinterface.qdoc
llewelld: Thanks, that is clearer; I've changed it.

import/popups/AuthDialog.qml
llewelld: I stole your comment from #115 to make things clearer. I agree with your idea to have some kind of configurable value here, but maybe it's something we could look at separately, using a reusable utility function (going by the comment in the gecko code, we should probably be localising the ellipsis, but then things start to get more complicated).

import/popups/PasswordManagerDialog.qml
llewelld: Thanks! I'm not sure why this wasn't generating an error, but I've put the string back in now at any rate.

import/popups/PopupOpener.qml
llewelld: Updated to use selector().

import/popups/PopupOpener.qml
llewelld: Thanks for spotting these. I've removed the ones I could see.

llewelld:
Rebased.

llewelld:
I've updated this PR to use qtTrIds generated direction from the gecko properties files using the `generate_property_translations.py` tool.

This should now be ready for review so I'm removing the WIP prefix.

llewelld:
I've added the commits for "JB#54829/OMP#JOLLA-221 - Find outstanding gecko strings for conversion to Qt id" on top of this one since it relates.

import/popups/AuthDialog.qml
llewelld: Now that the key ids and parameters are passed as a bundle to the UI, the trimming now happens in gecko again rather than the QML, so this bit of code has been removed.

llewelld:
Added sledges as a review for l10n support.
Added Tomi as a review in relation to the gecko-key-to-qt-id conversion.

import/popups/AuthPopupInterface.qml
```diff
@@ -13,6 +13,7 @@ import Sailfish.Silica 1.0
 
 UserPromptInterface {
     // inputs
+    property variant messageBundle
```
rainemak, line 16: Use var instead of variant as it's obsolete. Using variant may have serious performance implications.

Found this old commit:
https://github.com/sailfishos/sailfish-browser/commit/2eb40f46f92cfcd442556bac68e6b03561d05f18

[1]
https://doc.qt.io/qt-5/qml-variant.html

import/popups/PopupOpener.qml
```diff
@@ -34,7 +34,8 @@ Timer {
         "embed:auth":           "authPopup",
         "embed:permissions":  { "geolocation": "locationPermissionPopup" },
         "embed:webrtcrequest":  "webrtcPermissionPopup",
-        "embed:popupblocked":   "blockedTabPopup"
+        "embed:popupblocked":   "blockedTabPopup",
+        "embed:select":         "selectorPopup"
```
rainemak, line 38: How did you ended up resurrecting long time ago deprecated embed:select?

One reason would well be that it got dropped quite recently:
https://github.com/sailfishos/embedlite-components/commit/c0f7f03ef39db0bb915925284f7c696cf7370703

There's "embed:selectasync" in PickerOpener for handling \<select\> (single and multi).

doc/sailfish-webview-popups-popupprovider.qdoc
```diff
@@ -122,3 +122,11 @@
   The component specified must implement the
   \l{BlockedTabPopupInterface}{blocked tab popup} interface.
 */
+
+/*!
+  \qmlproperty var PopupProvider::selectorPopup
+  \brief The selector popup component descriptor
+
+  The component specified must implement the
+  \l{SelectorPopupInterface}{selector popup} interface.
```
rainemak, line 131: As commented, there's SingleSelectPage and MultiSelectDialog under pickers.

import/popups/AuthPopupInterface.qml
llewelld: Thanks for this. I've updated the variables to use `var` instead of `variant` (I wasn't aware of this, so useful for the future too).

import/popups/PopupOpener.qml
llewelld: I'll resolve this comment as per the discussion on the related changes in embedlite-components: https://github.com/sailfishos/embedlite-components/pull/5#discussion_r667088133

doc/sailfish-webview-popups-popupprovider.qdoc
llewelld: Similarly, I'll resolve this following the discussion on the related changes in embedlite-components: https://github.com/sailfishos/embedlite-components/pull/5#discussion_r667088133

chriadam (approved):

okodron (approved):

rainemak:
Rebase and resolve conflicts.

rainemak:
Rebased here https://github.com/sailfishos/sailfish-components-webview/pull/124

PR #118, opened 2021-07-02
[sailfish-components-webview] Fix for blocked popups. Fixes JB#54830 OMP#JOLLA-215

mkenttala:
WIP fix, or is there some better way to handle this?

### DIFF
--- a/import/controls/permissionmanager.cpp
+++ b/import/controls/permissionmanager.cpp
@@ -14,9 +14,12 @@
 
 PermissionManager::PermissionManager(QObject *parent) : QObject(parent) {
     SailfishOS::WebEngine *webEngine = SailfishOS::WebEngine::instance();
+    connect(webEngine, &SailfishOS::WebEngine::recvObserve, this, &PermissionManager::handleRecvObserve);
 
     webEngine->addObserver(QStringLiteral("embed:perms:all"));
     webEngine->addObserver(QStringLiteral("embed:perms:all-for-uri"));
+
+    sendRequest(QStringLiteral("get-all"));
 }
 
 void PermissionManager::add(const Permission &permission)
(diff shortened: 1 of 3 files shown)


rainemak:
Please update bug id 54803

import/popups/PopupOpener.qml
```diff
@@ -344,6 +344,22 @@ Timer {
...
 
     function blocked(data) {
+        switch (Controls.PermissionManager.popupCapability(data.host)) {
+            case Controls.PermissionManager.Unknown:
+            case Controls.PermissionManager.Prompt:
+                askPopup(data)
+                break
+            case Controls.PermissionManager.Allow:
+                //TODO: no confirmation?
+                askPopup(data)
```
rainemak, line 354: If popup is already allowed why to ask again?

import/controls/permissionmanager.cpp
```diff
@@ -26,11 +29,13 @@ void PermissionManager::add(const Permission &permission)
                 permission.m_type,
                 permission.m_capability,
                 permission.m_expireType);
+    sendRequest(QStringLiteral("get-all"));
```
rainemak, line 32: Rather than reading all back you should consider adding notify to the embedlite after successfully added. Similarly for removed case.

Bare minimum would be that this would call, "sendRequest(QStringLiteral("get-all-for-uri"), permission.m_host)";

See... PermissionModel::requestPermissions(const QString &host)

rainemak: On embedlite-components side and in ContentPermissionManager we could handle "perm-changed" and build and send updated list after changes are made.

You can find a browser component usage from here https://github.com/sailfishos/gecko-dev/blob/master/browser/components/preferences/sitePermissions.js#L97

So adding Services.obs.addObserver(this, "perm-changed") upon start and clearing it upon shutdown  Services.obs.removeObserver(this, "perm-changed").

mkenttala:
This turned out to be a lot easier to fix with just javascript: https://github.com/sailfishos/embedlite-components/pull/6

PR #119, opened 2021-07-06
[webview] Use flickable for each custom popup example. Fixes JB#54846 OMP#JOLLA-232

rainemak:


mkenttala (approved):

okodron (approved):

examples/custompopups/CustomAlertPopup.qml
```diff
@@ -20,42 +20,45 @@ AlertPopupInterface {
...
-            onClicked: { popup.accepted(); popup.visible = false }
+        SilicaFlickable {
+            id: flickable
+            anchors.fill: parent
+            contentHeight: content.height + Theme.paddingLarge
+
+            Column {
+                id: content
+                anchors.centerIn: parent
+                spacing: Theme.paddingLarge
```
llewelld, line 33: Without adding an explicit width, the Column will only grow to the largest element, which in this case is the button. It might be good to span the width of the parent by adding:
```
                width: parent.width - 2 * Theme.paddingLarge
```
Similarly for the other popups.

examples/custompopups/main.qml
```diff
@@ -53,57 +53,55 @@ ApplicationWindow {
...
-                    onClicked: { popup.accepted(); popup.visible = false }
-                }
-                Button {
-                    id: rejectBtn
-                    anchors.horizontalCenter: message.horizontalCenter
-                    anchors.top: acceptBtn.bottom
-                    anchors.topMargin: Theme.paddingLarge
-                    text: popup.cancelText
-                    onClicked: { popup.rejected(); popup.visible = false }
+                SilicaFlickable {
```
llewelld, line 58: A historical omission, but this file needs a copyright header.

llewelld (commented):
Just a couple of very small comments. It works much better now.

examples/custompopups/CustomAlertPopup.qml
rainemak: Even though this is not fullscreen popup, changed to "width: parent.width - 2 * Theme.horizontalPageMargin"

llewelld: Looks much better, thanks.

llewelld (approved):
LGTM also.

atatarov (approved):

PR #120, opened 2021-07-06
[sailfish-components-webview] Fix for input prompts. Fixes JB#54845 OMP#JOLLA-231

mkenttala:
The binging doesn't work for some reason

### DIFF
--- a/import/popups/PromptDialog.qml
+++ b/import/popups/PromptDialog.qml
@@ -18,7 +18,7 @@ Dialog {
     canAccept: input.text.length > 0
 
     property alias text: prompt.text
-    property alias value: prompt.value
+    property alias value: input.text
 
     property alias acceptText: prompt.acceptText
     property alias cancelText: prompt.cancelText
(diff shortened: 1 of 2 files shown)


import/popups/PromptDialog.qml
```diff
@@ -18,7 +18,7 @@ Dialog {
     canAccept: input.text.length > 0
 
     property alias text: prompt.text
-    property alias value: prompt.value
+    property alias value: input.text
```
rainemak, line 21: If "value: input.text" is not updated, then that should imply that prompt value that is send back to the engine [1] is not right either.

[1]
https://github.com/sailfishos/sailfish-components-webview/blob/master/import/popups/PopupOpener.qml#L189

mkenttala: That PromptDialog.value property is actually what's sent to the server, without this server always receives nothing from javascript prompt method.

rainemak: Yeap... and `CustomPromptPopup.qml` would need something like:
```
onAccepted: value = input.text
```
and removing that input.text binding.

rainemak: Maybe even:
```
Binding { target: popup; property: "value"; value: input.text }
```

Tried to rename "value" property to promptValue => works as well.

rainemak (approved):
Note nicest fixes but these will indeed fix this. Hence, approving.

atatarov (approved):

PR #121, opened 2021-07-27
[sailfish-components-webview] Prefill prompt input field with default value. Contributes to JB#52387 OMP#JOLLA-257

chriadam:
When passed a default value to be prefilled into the prompt input
field, we should set the input field's text value to that value.

In order to prevent two-way-binding problem, we need to pass this
default value as a separate input property, rather than attempting
to bind it to the (output) value property.

### DIFF
--- a/import/popups/PromptPopupInterface.qml
+++ b/import/popups/PromptPopupInterface.qml
@@ -14,6 +14,7 @@ import Sailfish.Silica 1.0
 UserPromptInterface {
     // inputs
     property string text
+    property string defaultValue
 
     // outputs
     property string value
--- a/examples/custompopups/CustomPromptPopup.qml
+++ b/examples/custompopups/CustomPromptPopup.qml
@@ -21,8 +21,7 @@ PromptPopupInterface {
     cancelText: "Cancel"
 
     preventDialogsValue: toggle.checked
-
-    onAccepted: value = input.text
+    value: input.text
 
     Rectangle {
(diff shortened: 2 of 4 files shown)


import/popups/PromptPopupInterface.qml
```diff
@@ -14,6 +14,7 @@ import Sailfish.Silica 1.0
 UserPromptInterface {
     // inputs
     property string text
+    property string defaultValue
```
chriadam, line 17: This is an addition to the interface API.  Is this allowed, at this stage, or are these APIs supposed to be considered stable already?

atatarov (approved):

import/popups/PromptPopupInterface.qml
llewelld: Just sumarising an internal discussion on this point (@atatarov, please correct me if I'm mischaracterising it): 
@llewelld: "What's OMP's position on this? Are we still able to adapt the API?"
@atatarov: "I think it's ok"

okodron: It's ok to change such API at this stage.

okodron (approved):

import/popups/PromptPopupInterface.qml
rainemak: Marking as resolved.

rainemak (approved):

PR #122, opened 2021-08-04
[sailfish-components-webview] Prevent binding loop in prompt dialog. Contributes to JB#55000

chriadam:

PR #123, opened 2021-08-04
[sailfish-components-webview] Appropriately detect preventAdditionalDialogs prefill value. Contributes to JB#55035

chriadam:
The prefill value is passed as "checked" not as "value", according
to 0279c8930f1dcac9076503db63a1ee94b7ca9e0a in embedlite-components.
(Note also the typo: "addional" in embedlite-components.)

### DIFF
--- a/import/popups/PopupOpener.qml
+++ b/import/popups/PopupOpener.qml
@@ -69,6 +69,18 @@ Timer {
         return null
     }
 
+    function getCheckboxValue(checkbox) {
+        if (checkbox) {
+            // Check for undefined / null
+            if (checkbox.checked !== null && checkbox.checked !== undefined) {
+                return checkbox.checked
+            } else if (checkbox.value !== null && checkbox.value !== undefined) {
+                return checkbox.value
+            }
+        }
+        return false
+    }
+
     // Returns true if message is handled.
     function message(topic, data) {
         if (!handlesMessage(topic)) {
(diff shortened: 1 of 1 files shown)


chriadam:
Note: needs to be retargeted to master after https://github.com/sailfishos/sailfish-components-webview/pull/122 is merged

okodron (approved):

import/popups/PopupOpener.qml
```diff
@@ -69,6 +69,16 @@ Timer {
         return null
     }
 
+    function getCheckboxValue(checkbox) {
+        return (checkbox != null && checkbox != undefined)
```
rainemak, line 73: if, else if, else if, else block would work better with this kind of deeply nested ternary operator.

This is type driven checking... Worth adding strict equal comparison.

rainemak: Maybe strictly comparing against false and true for checked.checked and checkbox.value would rendered a bit cleaner block.

```
if (checkbox) {
    // Check for undefined / null
    if (checkbox.checked === true || checkbox.checked == false) {
        return checkbox.checked
    } else if (checkbox.value === true || checkbox.value ===false) {
        return checkbox.value
    }
}

return false
```

or 

```
if (checkbox) {
    // Check for undefined / null
    if (checkbox.checked !== null && checkbox.checked !== undefined) {
        return checkbox.checked
    } else if (checkbox.value !== null && checkbox.value !== undefined) {
        return checkbox.value
    }
}

return false
```

I think decision should be based on the possible checked / value types.

okodron: Yeah, that's better.

chriadam: Indeed, thanks, done

rainemak (approved):
Didn't test on device but code looks good. Hence, approving.

chriadam:
Rebased.

PR #125, opened 2021-09-07
[sailfish-webview] Advise PkgConfig usage properly. Fixes JB#55427

vigejolla:


doc/index.qdoc
```diff
@@ -17,11 +18,17 @@ Mozilla gecko.
...
+the spec file you should add line:
+
+\code
+BuildRequires:  pkgconfig(qt5embedwidget)
+\endcode
+
+And in your pro file you should add line:
 
 \code
-CONFIG += link_pkgconfig
```
rainemak, line 24: As you mentioned, this is needed if you do not sailfishapp. Good to promote usage of that as well.

rainemak: At the same having a CONFIG twice doesn't really harm.

vigejolla: It does harm - it doesn't work. You can't have both sailfishapp and link_pkgconfig there.

vigejolla: But I'll rewrite this section so that it mentions both

rainemak: Thank @vigejolla :+1:

rainemak (approved):

PR #127, opened 2021-10-04
[sailfish-webview] Adjust text selection markers by scrollable offset. Fixes JB#55606 OMP#JOLLA-400

rainemak:
Not entirely sure what has changed in embedlite-components or gecko but
now coordinates that TextSelectionController receives from the
embedlite-components (Content:SelectionRange) need to adjusted by
scroll offset in addition to resolution (zoom).

Reason for this could be as simple as Util.translateToTopLevelWindow
(embedlite-components Util.js), returning offset differently. That would
imply that initial context menu position would be different as
well (_processPopupNode) and context menu request changing to selection start
(Browser:SelectionStart) would pass the same different coordinates back
to the embedlite-components.

Engine side highlight was ok already before only user interface side markers
(QML items) were at wrong position.

### DIFF
--- a/import/controls/TextSelectionController.qml
+++ b/import/controls/TextSelectionController.qml
@@ -49,8 +49,9 @@ MouseArea {
         var state = data.src
 
         // Start marker
-        start.fixedX = (data.start.xPos * resolution) - start.width
-        start.fixedY = data.start.yPos * resolution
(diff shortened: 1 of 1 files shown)


mkenttala (approved):

chriadam:
So, is the initial context menu position incorrect now by this offset amount also?  If so, is it better to fix Util.js rather than work around the change here?  Or am I misunderstanding?

PR #129, opened 2021-10-25
[sailfish-webview] Introduce webview test application. Contributes to JB#56037

paveltkv:


atatarov (approved):

okodron (approved):

examples/webview-test-app/main.qml
```diff
@@ -0,0 +1,272 @@
...
+
+                    active = value
+                }
+
+                anchors.fill: parent
+                active: false
+                sourceComponent: webViewComponent
+
+                onItemChanged: {
+                    if (active) {
```
rainemak, line 226:
```suggestion
                    if (item) {
```

examples/webview-test-app/main.qml
```diff
@@ -0,0 +1,272 @@
...
+                        console.log("page loading: " + webPageAddress)
+                        item.load(webPageAddress)
+                    }
+                }
+            }
+
+            Rectangle {
+                id: loadStatus
```
rainemak, line 234: Unused

examples/webview-test-app/main.qml
```diff
@@ -0,0 +1,272 @@
...
+
+                color: "grey"
+                anchors.bottom: parent.bottom
+                anchors.left: parent.left
+                anchors.margins: 22
```
rainemak, line 239: It's good habit to group anchors when possible. So like:

```
anchors {
    bottom: parent.bottom
    left: parent.left
    margins: 22
}
```

examples/webview-test-app/SpinBox.qml
```diff
@@ -0,0 +1,64 @@
...
+    property real stepSize: 0.1
+
+    implicitHeight: Theme.itemSizeMedium
+    implicitWidth: 200
+
+    onValueChanged: {
+        value = Math.min( Math.max(minimumValue, value), maximumValue )
+    }
+
+    IconButton {
```
rainemak, line 19: For this kind of a case Row work nicely. By using Row you don't need to fiddle with z-index and sibling anchors.

```
    Row {
        width: parent.width
        anchors.verticalCenter: parent.verticalCenter
        IconButton {
            id: buttonMinus
            width: height
            anchors.verticalCenter: textField.verticalCenter
            icon.source: "image://theme/icon-video-overlay-play"
            rotation: 180
            onClicked: root.value -= root.stepSize
        }

        TextField {
            id: textField

            inputMethodHints: Qt.ImhFormattedNumbersOnly
            placeholderText: label
            text: root.value.toFixed(2)
            horizontalAlignment: TextInput.AlignHCenter
            width: parent.width - 2*buttonMinus.width
            anchors.verticalCenter: parent.verticalCenter

            EnterKey.iconSource: "image://theme/icon-m-enter-next"
            EnterKey.onClicked: {
                focus = false;
                root.value = text;
            }

            onFocusChanged: {
                if (!focus) {
                    root.value = text;
                }
            }
        }

        IconButton {
            height: buttonMinus.height
            width: buttonMinus.height
            anchors.verticalCenter: textField.verticalCenter
            icon.source: "image://theme/icon-video-overlay-play"
            onClicked: root.value += root.stepSize
        }
    }
```

examples/webview-test-app/SpinBox.qml
```diff
@@ -0,0 +1,64 @@
...
+    property real maximumValue: Infinity
+    property real value: 0.0
+    property real stepSize: 0.1
+
+    implicitHeight: Theme.itemSizeMedium
+    implicitWidth: 200
+
+    onValueChanged: {
+        value = Math.min( Math.max(minimumValue, value), maximumValue )
+    }
```
rainemak, line 17: With one liner signal handlers it's not mandatory to use curly brackets. Following would be equally good:

```
onValueChanged: value = Math.min(Math.max(minimumValue, value), maximumValue)

File: examples/webview-test-app/main.qml [hunk 6]
```diff
@@ -0,0 +1,272 @@
...
+                    onLoadProgressChanged: {
+                        webViewPage.webViewLoadProgress = loadProgress
+                    }
+                }
+            }
+
+            Loader {
+                id: webViewLoader
+
+                function setWebViewActive(value) {
```
rainemak, line 207: Is it not needed rather even a bit misleading. I'd rather just use ```webViewLoader.active``` where needed.

For logging purposes you could add

```
onActiveChanged: {
    if (value) {
        console.log("webview creation")
    } else {
        console.log("webview destruction")
    }
}
```

Changing a value of a property to the same that there was should not trigger change. That is the normal way how properties should be written in C++. This implies that...

```
webViewLoader.active = true
webViewLoader.active = true
```

Would log "webview creation" once given that previous active value was false.

rainemak (commented):
Only minor remarks. Please give a look.

rainemak (approved):

PR #131, opened 2021-11-01
Act on private browsing when requesting permissions. JB#55925 OMP#JOLLA-445

llewelld:
A `privateBrowsing` flag is now passed with the "embed:permissions" message when the browser wants to request a permission of the user (a location permission specifically).

This change disables the option to store the permission when private browsing is enabled, to avoid leaking location data.

### DIFF
--- a/import/popups/PopupOpener.qml
+++ b/import/popups/PopupOpener.qml
@@ -342,7 +342,8 @@ Timer {
     // Open permissions dialog
     function permissions(data) {
         var props = {
-            "host": data.host
+            "host": data.host,
+            "privateBrowsing": data.privateBrowsing || true
         }
         var acceptFn = function(popup) {
             _popupObject = null
--- a/doc/sailfish-webview-popups-locationpopupinterface.qdoc
+++ b/doc/sailfish-webview-popups-locationpopupinterface.qdoc
@@ -27,6 +27,19 @@
 */
 
 /*!
+  \qmlproperty bool LocationPopupInterface::privateBrowsing
+  \brief Will be true if the user is in private browsing mode
+
+  A custom implementation should use the value of this property to determine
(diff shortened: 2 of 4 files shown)


llewelld:
Requires https://github.com/sailfishos/embedlite-components/pull/65

For info, with these PRs the behaviour of the Location Permission is now as follows:

In normal browsing mode:
  1. Accepting or Denying the permission temporarily will last until the browser is restarted. The setting will apply to private browsing mode as well.
  2. Accepting or Denying the permission permanently will last until the user clears the permissions via the settings. The setting will apply to private browsing mode as well.

In private browsing mode:
  1. Accepting or Denying the permission temporarily will be applied only once. If the site requests access again the user will be asked again. This has no effect on normal mode because nothing is stored.
  2. There is no option to Accept or Deny permanently.

import/popups/PopupOpener.qml
```diff
@@ -342,7 +342,8 @@ Timer {
     // Open permissions dialog
     function permissions(data) {
         var props = {
-            "host": data.host
+            "host": data.host,
+            "privateBrowsing": data.privateBrowsing || false
```
rainemak, line 346: Does data of "embed:permissions" carry now always privateBrowsing, doesn't it. In order to be in sync with https://github.com/sailfishos/embedlite-components/pull/65/files#diff-e85ea15f29d567f030ec05fd328e2a6c011e3713d016630b1e36100c5d22da49R39 should this be ```"privateBrowsing": data.privateBrowsing || true```

llewelld: A good point, and nicely spotted, thanks. I've updated the code.

okodron (approved):

rainemak (approved):

llewelld:
Rebased.

PR #133, opened 2021-11-05
Ensure geo permissions can be saved in normal browsing. Fixes JB#56207 OMP#JOLLA-509

llewelld:
A (rather embarrassing; sorry!) logic error meant that the permission dialogue thought it was always in private browsing mode, preventing the user from saving their geolocation permission setting.

Since the `privateBrowsing` entry will always be included in the message data, there should be no need to guard it anyway.


rainemak (approved):

okodron (approved):

llewelld:
Rebased.

PR #140, opened 2021-12-07
[sailfish-components-webview] Only populate permissions model with known permissions. Contributes to JB#56450 OMP#JOLLA-550

chriadam:
If a permission isn't one of the permissions we are aware of (and
have an icon for) don't include it in the permissions model which
exposes it to QML.

Contributes to JB#56450

### DIFF
--- a/import/controls/permissionmodel.cpp
+++ b/import/controls/permissionmodel.cpp
@@ -134,9 +134,20 @@ void PermissionModel::handleRecvObserve(const QString &message, const QVariant &
 
 void PermissionModel::setPermissionList(const QVariantList &data)
 {
+    static const QStringList knownPermissions {
+        QStringLiteral("geolocation"),
+        QStringLiteral("cookie"),
+        QStringLiteral("popup"),
+        QStringLiteral("camera"),
+        QStringLiteral("microphone"),
+    };
+
     QList<Permission> permissions;
     for (const auto &iter : data) {
         QVariantMap varMap = iter.toMap();
+        if (!knownPermissions.contains(varMap.value("type").toString()))
+            continue;
+
         permissions.append(Permission(varMap.value("uri").toString(),
                                       varMap.value("type").toString(),
                                       PermissionManager::intToCapability(varMap.value("capability").toInt()),


chriadam:
This is still WIP.  Need to confirm that this approach is valid, and that we cannot enumerate the supported permissions programmatically.

llewelld (approved):
Personally I think this approach makes sense. Since we're not providing icons for permissions outside this set, they have to be filtered at some point, and this seem a reasonable place. We're also making similar assumptions in [other parts of the UI code](https://github.com/sailfishos/sailfish-browser/blob/3eacd32afdc90bf214aaaed203d5d9b0df4cfb91/apps/browser/qml/pages/PermissionPage.qml#L57), so while it'd be nice if this could be queried from somewhere, this information should then ideally be propagated to other parts of the UI code too.

llewelld:
I should add that I tested this and found that it addressed the problem: the "Site permissions" button now appears on the sites where it was missing before (at least for the few I tested).

llewelld:
One small point: the commit title needs updating to include `OMP#JOLLA-550`.

chriadam:
Thanks, updated commit message with bug tag.

Do you have a suggestion for:
a) whether we should centralise the "known permission types"
b) how to offer that as API which can be queries in the appropriate places

Or, do you think we should leave this as-is for now, to avoid unnecessary complexity?

llewelld:
> a) whether we should centralise the "known permission types"
> b) how to offer that as API which can be queries in the appropriate places

I'm not sure. Off the top of my head, it probably doesn't make sense to change anything in the privileged javascript (gecko + embedlite-components) layers where they're all handled as strings already, since this approach comes from upstream.

That leaves the QML layer, where it might make more sense to provide an enum (maybe exposed through the [PermissionModel](https://github.com/sailfishos/sailfish-components-webview/blob/211fe78f5a5a7a031f2d417df850b1d8cda5e3d8/import/controls/permissionmodel.h#L55) here?) instead of using strings.

If there's a very specific set of permissions that the UI is willing/able to handle (which seems to be the case, and also seems unlikely to change), then using an enum in the QML layer would make sense in my opinion.

> Or, do you think we should leave this as-is for now, to avoid unnecessary complexity?

Unless it's a very quick task, I'd personally be inclined to create another task for it. In my opinion your PR here already fixes the issue for the user.

Eventually we may need to provide an interface to the permissions as part of the [PopupInterface](https://github.com/sailfishos/sailfish-components-webview/blob/211fe78f5a5a7a031f2d417df850b1d8cda5e3d8/import/popups/PopupOpener.qml#L139). So it might be worth thinking about whether it would be better to provide this using an enum or a string. Maybe that could also fall under the task.

paveltkv (approved):

atatarov (approved):

okodron (approved):

PR #141, opened 2021-12-13
[sailfish-components-webview] Fix build break. Contributes to JB#56631 OMP#JOLLA-569

chriadam:

PR #142, opened 2021-12-13
[webview] Align to ViewCreator API changes. JB#56631 OMP#JOLLA-569

rainemak:
Browsing context cannot be optional for sub windows.

### DIFF
--- a/import/webview/rawwebview.cpp
+++ b/import/webview/rawwebview.cpp
@@ -37,7 +37,7 @@ public:
     ViewCreator();
     ~ViewCreator();
 
-    quint32 createView(const quint32 &parentId) override;
+    quint32 createView(const quint32 &parentId, const uintptr_t &parentBrowsingContext) override;
 
     static std::shared_ptr<ViewCreator> instance();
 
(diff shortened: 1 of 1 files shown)


rainemak:
This PR supersedes https://github.com/sailfishos/sailfish-components-webview/pull/141

okodron (approved):

mkenttala (approved):

import/webview/rawwebview.h
```diff
@@ -52,7 +52,7 @@ class RawWebView : public QuickMozView
     void footerMarginChanged();
     void contentOrientationChanged(Qt::ScreenOrientation orientation);
     void acceptTouchEventsChanged();
-    void openUrlInNewWindow();
+    void openUrlInNewWindow(const quint32 &parentId, const uintptr_t &parentBrowsingContext);
```
chriadam, line 55: is this signal usually handled by C++ code or by QML code?  If the latter, it might have some issues with the second parameter.

chriadam:
One comment (which might not be valid anyway), aside from that LGTM.

import/webview/rawwebview.h
rainemak: This signal is completely wrong. Let's drop all arguments. We cannot really do anything meaningful with this API at the moment. That's also the reason why requested this to internal whilst documenting as this won't work well without parentId either.

rainemak: Did somewhat similar cleanup here: https://github.com/sailfishos/qtmozembed/pull/31/files#diff-334faccf9531e149b1c3bd2583bf513303a0b44e54a3b8c84b5a3a7037abf3fbR87

rainemak:
> One comment (which might not be valid anyway), aside from that LGTM.

Currently API is marked as internal (highlighted it in previous comment). Mismatches between parent-child browsing contexts will assert inside the engine. Only way to bypass on QML side window.open to somewhat work is to break parent-child relationship between parent&child webview => i.e. treat parentId always is zero. This of course have implications like parent web content cannot close created child not can child close itself.

This implies that for QML we need at some point better API.

PR #145, opened 2021-12-17
Document overloaded setPreference C++ method

rainemak:
From QML point of view we can have setPreference documented as one method with one optional argument.

### DIFF
--- a/doc/sailfish-webview-webenginesettings.cpp
+++ b/doc/sailfish-webview-webenginesettings.cpp
@@ -220,12 +220,8 @@ void SailfishOS::WebEngineSettings::enableLowPrecisionBuffers(bool enabled);
 
     This is a low-level API to set engine preferences. Higher-level calls, such
     as those provided by \l WebEngineSettings,
-    \l {SailfishOS::WebEngineSettings}{WebEngineSettings},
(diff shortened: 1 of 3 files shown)


rainemak:
See also https://github.com/sailfishos/qtmozembed/pull/34

okodron (approved):

doc/sailfish-webview-webenginesettings.cpp
```diff
@@ -212,4 +229,15 @@ void SailfishOS::WebEngineSettings::enableLowPrecisionBuffers(bool enabled);
 */
 void SailfishOS::WebEngineSettings::setPreference(const QString &key, const QVariant &value);
```
llewelld, line 230: The changes here are fine, but I think line 206/223 of the original description of `setPreference` is redundant (since the same link to `WebEngineSettings` is already provided on the line above). This might be a good opportunity to remove it.

The link to the MDM docs is also unfortunately broken, so probably not worth keeping.

llewelld (approved):

doc/sailfish-webview-webenginesettings.cpp
rainemak: Added a commit on top that cleans them up.

PR #148, opened 2021-12-23
Disable wasm_baselinejit on lower memory devices. JB#56635 OMP#JOLLA-572

llewelld:
WebAssembly, and in particular the baselinejit, seems to have a heavy
memory footprint, which can cause the browser to crash on lower memory
devices (approx 2Gb or lower).

This change automatically disables wasm_baselinejit on devices with less
than 2.5Gb of RAM.

### DIFF
--- a/lib/logging.cpp
+++ b/lib/logging.cpp
@@ -1,0 +1,15 @@
+/****************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "logging.h"
+
+Q_LOGGING_CATEGORY(lcWebviewLog, "org.sailfishos.webview", QtWarningMsg)
+Q_LOGGING_CATEGORY(lcWebengineLog, "org.sailfishos.webengine", QtWarningMsg)
+Q_LOGGING_CATEGORY(lcWebenginesettingsLog, "org.sailfishos.webenginesettings", QtWarningMsg)
--- a/lib/logging.h
+++ b/lib/logging.h
@@ -1,0 +1,20 @@
+/****************************************************************************
+**
+** Copyright (c) 2021 Open Mobile Platform LLC.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#ifndef SAILFISHOS_LOGGING_H
+#define SAILFISHOS_LOGGING_H
+
+#include <QLoggingCategory>
+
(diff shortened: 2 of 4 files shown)


lib/webenginesettings.cpp
```diff
@@ -102,6 +109,14 @@ void SailfishOS::WebEngineSettings::initialize()
     // Make long press timeout equal to the one in Qt
     engineSettings->setPreference(QStringLiteral("ui.click_hold_context_menus.delay"), QVariant(PressAndHoldDelay));
 
+    // Disable wasm_baselinejit to avoid crashes on lower memory devices; see JB#56635
+    quint64 totalMemory = getTotalMemory();
+    if (totalMemory < (2.5 * 1024 * 1024 * 1024)) {
+        // Devices with roughly 2Gb and lower
+        qDebug() << "Lower memory: disabling wasm_baselinejit";
```
rainemak, line 116: sailfish-components-webview deserves a logging categories. This would be info kind of a thing. Created a new task about logging categories (JB#56816).

rainemak: If you're aiming to keep this, there should "#include <QDebug>"

llewelld: I've added some logging categories, as an extra commit on top.

lib/logging.cpp
```diff
@@ -0,0 +1,16 @@
+/****************************************************************************
+**
+** Copyright (C) 2021 Jolla Ltd.
```
rainemak, line 3: This can be dropped

lib/logging.h
```diff
@@ -0,0 +1,21 @@
+/****************************************************************************
+**
+** Copyright (C) 2021 Jolla Ltd.
```
rainemak, line 3: Same

lib/logging.cpp
```diff
@@ -0,0 +1,16 @@
...
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include "logging.h"
+
+Q_LOGGING_CATEGORY(lcWebviewLog, "org.sailfishos.browser.webview", QtWarningMsg)
```
rainemak, line 14: "org.sailfishos.webview" instead

lib/logging.cpp
```diff
@@ -0,0 +1,16 @@
...
+Q_LOGGING_CATEGORY(lcWebengineLog, "org.sailfishos.browser.webengine", QtWarningMsg)
+Q_LOGGING_CATEGORY(lcWebenginesettingsLog, "org.sailfishos.browser.webenginesettings", QtWarningMsg)
```
rainemak, line 16: Drop browser from both

lib/logging.cpp
llewelld: Dropped.

lib/logging.h
llewelld: Dropped.

lib/logging.cpp
llewelld: Changed.

lib/logging.cpp
llewelld: Changed.

rainemak (approved):

okodron (approved):

PR #152, opened 2022-01-31
[components-webview] Fix click handling after selection. Contributes to JB#56311

mkenttala:


import/controls/TextSelectionController.qml
```diff
@@ -158,8 +160,20 @@ MouseArea {
         }
     }
 
+    function selectionRect() {
+        var selectionX = Math.min(start.x, end.x)
+        var selectionY = Math.min(start.y, end.y)
+        var selectionWidth = Math.max(start.x, end.x) - selectionX + end.width
+        var selectionHeight = Math.max(start.y, end.y) - selectionY + end.height
+        var rv = Qt.rect(selectionX, selectionY, selectionWidth, selectionHeight)
+        return rv
```
rainemak, line 169: Not quite sure why but initial size is off after entering to the text selection mode. After moving a bit left or right handle, selection rectangle updates to somewhat meaningful.

mkenttala:
Modified to use separate rects for the start and end handles. Also made the mousearea smaller so it's a lot easier to hit the quote button in sailfish forum. Because of the animation in the handles I wasn't able to use mousearea directly so the rects are calculated whenever animations stop.

import/controls/TextSelectionHandle.qml
```diff
@@ -53,12 +54,15 @@ Image {
         } else {
             contentItem.sendAsyncMessage("Browser:SelectionMoveEnd", selectionController.getMarkerBaseMessage(markerTag))
             moving = false
+            rect = Qt.rect(x - Theme.paddingLarge, y, width + 2 * Theme.paddingLarge, height + Theme.paddingLarge)
```
rainemak, line 57:
```suggestion
            rect = Qt.rect(mouseArea.x, mouseArea.y, mouseArea.width, mouseArea.height)
```

import/controls/TextSelectionHandle.qml
```diff
@@ -73,9 +77,9 @@ Image {
 
         anchors {
             fill: parent
```
rainemak, line 79: Width and height above are unnecessary as there's anchors.fill.

import/controls/TextSelectionHandle.qml
mkenttala: That was my first approach but it doesn't work, mouseArea only changes later and even binding to it doesn't seem to work.

import/controls/TextSelectionHandle.qml
```diff
@@ -41,6 +41,7 @@ Image {
     property alias targetPositionAnimation: targetPositionAnimation
 
     property string markerTag
+    property rect rect: Qt.rect(0, 0, 0, 0)
```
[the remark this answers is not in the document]
mkenttala: Nope, for some reason binding to mouseArea is not working.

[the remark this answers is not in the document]
rainemak: Answering myself => no it wouldn't. Updating single rect property will not trigger update for the property itself.

rainemak, line 44: I'd avoid calling this rect, something like ```touchArea```.

import/controls/TextSelectionHandle.qml
rainemak: My point is that now you have style / behaviour related values scattered here and there. 

Given no animators are used even following would be a bit cleaner:
```
    onXChanged: {
        touchArea = Qt.rect(x - Theme.paddingLarge, y, width + 2 * Theme.paddingLarge, height + Theme.paddingLarge)
    }
    onYChanged: {
        touchArea = Qt.rect(x - Theme.paddingLarge, y, width + 2 * Theme.paddingLarge, height + Theme.paddingLarge)
    }
```
Then you'd not need to care about animations and only special case would be visible false and clearing touchArea there.

rainemak (changes requested):
Few minor remarks added.

import/controls/TextSelectionHandle.qml
mkenttala: Ok, changed to that. A bit more stress for the cpu doesn't seem to affect performance.

rainemak (approved):
Looks good now

import/controls/TextSelectionHandle.qml
rainemak: Yeap, that's what I checked as well.

PR #159, opened 2022-05-05
[webview-docs] Elaborate on using local content with CORS protection. Fixes JB#57956

llewelld:
Adds a description to the docs of three different approaches that can be used for serving local content using a WebView, in order to avoid Cross Origin Resource Sharing protection issues.

### DIFF
--- a/doc/index.qdoc
+++ b/doc/index.qdoc
@@ -1,7 +1,7 @@
 /****************************************************************************************
 **
 ** Copyright (c) 2021 Open Mobile Platform LLC
-** Copyright (c) 2021 Jolla Ltd.
+** Copyright (c) 2021 - 2022 Jolla Ltd.
 ** All rights reserved.
 **
 ****************************************************************************************/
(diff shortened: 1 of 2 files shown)


llewelld:
@poetaster: it'd be good to get your thoughts on this.

poetaster:
> @poetaster: it'd be good to get your thoughts on this.

If just read through it and, as usual, clear and comprehensive! This will be of utility to many frustrated devs :)

doc/index.qdoc
```diff
@@ -115,34 +116,143 @@ WebView currently supports customization via the following interfaces:
...
+
+\section3 Option 1. Don't serve the files locally at all
+
+In many cases, while it can initially seem sensible to serve local files,
+you may find you can get the same or better capabilities by serving a
+remote version of your app instead.
+
+Many Web Apps will use caching, \l {https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API}{Web Storage}
+and similar technologies so that they can operate fully even when your
+phone is offline. The app will be served from the remote server on first
```
rainemak, line 155: so that they can operate fully even when offline.

Simpler not to mention "phone".

rainemak (approved):
Added a minor comment. Approving already as this looks like a good improvement / clarification.

doc/index.qdoc
llewelld: Yeah, that's better. Thank you.

PR #160, opened 2022-06-20
[webview] Guard programmatic preference change with marker file. Fixes JB#58269

rainemak:
Signed-off-by: Raine Makelainen <raine.makelainen@jolla.com>


lib/webenginesettings.cpp
```diff
@@ -178,6 +186,9 @@ void SailfishOS::WebEngineSettings::initialize()
                                   QVariant::fromValue<bool>(true));
```
llewelld, line 186: This comment relates to lines 163-172 (but it won't allow me to put a comment there).

I've not tried this out, but my understanding of the code is that the config values will only be forcefully set on first execution; after which they act like normal config values.

Most of the values are static and never change (except between devices). However the locale may change between browser runs. So I'm thinking maybe the code to set the `Accept-Language` header should sit outside the marker file guard?

llewelld (commented):
A nice solution, just one comment.

lib/webenginesettings.cpp
rainemak: Good catch, will move `Accept-Language` so that it will be executed always

rainemak: Fixed

llewelld (approved):
This LGTM. I tested using `layout.css.devPixelsPerPx` as described in [archon's original forum comment](https://forum.sailfishos.org/t/browser-configuration-does-not-store-integer-values/1017/6) and the behaviour now works as expected. Changing it in `about:config` sticks between browser restarts, unless the `~/.local/share/org.sailfishos/browser/__PREFS_WRITTEN__` file is removed. Deleting `__PREFS_WRITTEN__` causes the value to get reset on next restart.

PR #162, opened 2022-12-01
[webview] Correctly update viewportHeight on orientation change. Fixes JB#59065

llewelld:
When the device orientation changes the WebView height must be updated to match. Previously this was based on the screen height and width, but didn't take into account the possibility of non-full-screen use. Commit df9ae3ced8e2a7c tried to fix this by switching screen height for widget height.

However the fix didn't consider the fact that the height and width are already reversed on orientation change. The result was that in portrait mode the wrong height was set.

This change attempts to address this by always using the widget height.

### DIFF
--- a/import/webview/WebView.qml
+++ b/import/webview/WebView.qml
@@ -65,9 +65,7 @@ RawWebView {
             || _appActive && (webViewPage.status === PageStatus.Deactivating)
     _acceptTouchEvents: !textSelectionActive
 
-    viewportHeight: webViewPage
-            ? ((webViewPage.orientation & Orientation.PortraitMask) ? height : width)
-            : undefined
+    viewportHeight: webViewPage ? height : undefined
 
     orientation: {
         switch (_pageOrientation) {


llewelld:
@poetaster: I tested this change with your [webview-screensize](https://github.com/poetaster/webview-screensize) test tool, with positive results. But would you be able to give it some more robust testing with your other apps please?

I can supply you with a package to install, but it might be easier just to edit the `/usr/lib64/qt5/qml/Sailfish/WebView/WebView.qml` file on your device directly for testing (but please be aware this may cause breakage, so please keep a backup).

poetaster:
Ok, this makes scaling behaviour work properly. For instance, in mah-solitaire ... I just did the quick on device test. 

Sadly it also demonstrates that something else is broken on the volla/gs290/gs5 side of things. But, I can play with this warpper and see if I get any the wiser ;) Thanks!

llewelld:
Thanks for testing this @poetaster. I'm glad it fixes at least this issue. The Volla/gs290/gs5 issue is harder for me to contribute to as I don't have access to the devices, but if there's something you think I might be able to help with, please ask.

Thaodan (approved):
LGTM, the changes makes sense.

PR #165, opened 2023-09-06
Allow link ContextMenu options for image links. Add support for copying image URL in ContextMenu.

mlehtima:


rainemak:
Let's do some additions so that there would be also title visible when long pressing an image. At the same time we shall check that 2nd label wouldn't be the same as the title. By having title visible end user would have a bit more context for *Open link in new tab*, *Share*, *Copy Link*, *Save link*. *Share* would become a bit misleading as you wouldn't have context regarding what you are sharing. Is link a link or an image that you're sharing but I think we can live with that.

Further, noticed that with maximum number of items visible landscape context menu won't fit. Needs a bit of layout refactoring and flickable. @mlehtima if you don't mind, I could work a bit on the layout issues and push a commit to your branch.

rainemak (approved):
Changes that you did @mlehtima look good.

mlehtima:
Changes by @rainemak look good.

Tomin1 (approved):

thigg:
If the title is shown, finally all of xkcd can be read <3

rainemak:
> If the title is shown, finally all of xkcd can be read <3

Created a small PR https://github.com/sailfishos/embedlite-components/pull/95

Especially xkcd titles are very long. Thus, we should either calculate font size or just reduce font size. However, that's another PR.

rainemak:
Max 4 lines and smaller font: https://github.com/sailfishos/sailfish-components-webview/pull/166

Screenshot

PR #168, opened 2024-07-12
[sailfish-webview] Support external flag. JB#61045

rainemak:


llewelld (approved):
This is great; thank you @rainemak. I somehow forgot to push these changes to a remote branch, apologies for that. I'll create a separate PR with the other changes I've had to make (but which I think will need adjustment).

In the meantime, this looks good to me.

PR #169, opened 2024-07-19
Changes needed for ESR 91

llewelld:
These are the further changes I've needed to make to get the WebView working with the latest ESR 91 build. This adds two commits on top of  #168.

The first commit is to set the preferences as needed.

The second commit is to revert to the previous ofono import. I expect this second commit is only needed because of the SDK version I'm using (or maybe because I wasn't on devel?) and so almost certainly needs to be removed. I'm including it for completeness.

The first commit might be unnecessary as well?

### DIFF
--- a/lib/webenginesettings.cpp
+++ b/lib/webenginesettings.cpp
@@ -110,6 +110,12 @@ void SailfishOS::WebEngineSettings::initialize()
     engineSettings->setPreference(QStringLiteral("intl.accept_languages"),
                                   QVariant::fromValue<QString>(langs));
 
+    // Ensure the renderer is configured correctly
+    engineSettings->setPreference(QStringLiteral("gfx.webrender.force-disabled"),
+                                  QVariant(true));
+    engineSettings->setPreference(QStringLiteral("embedlite.compositor.external_gl_context"),
+                                  QVariant(false));
+
     Silica::Theme *silicaTheme = Silica::Theme::instance();
 
     // Notify gecko when the ambience switches between light and dark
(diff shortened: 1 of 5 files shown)


lib/webenginesettings.cpp
```diff
@@ -110,17 +127,32 @@ void SailfishOS::WebEngineSettings::initialize()
     engineSettings->setPreference(QStringLiteral("intl.accept_languages"),
                                   QVariant::fromValue<QString>(langs));
 
+    // Ensure the renderer is configured correctly
+    engineSettings->setPreference(QStringLiteral("gfx.webrender.force-disabled"),
+                                  QVariant(true));
+    engineSettings->setPreference(QStringLiteral("embedlite.compositor.external_gl_context"),
+                                  QVariant(false));
```
rainemak, line 134: I think embedding.js (embedlite omni.ja) should default to false and browser set it true.

lib/webenginesettings.cpp
```diff
@@ -110,17 +127,32 @@ void SailfishOS::WebEngineSettings::initialize()
     engineSettings->setPreference(QStringLiteral("intl.accept_languages"),
                                   QVariant::fromValue<QString>(langs));
 
+    // Ensure the renderer is configured correctly
+    engineSettings->setPreference(QStringLiteral("gfx.webrender.force-disabled"),
+                                  QVariant(true));
```
rainemak, line 132: This could go to embedding.js to the gecko side.

lib/webenginesettings.cpp
llewelld: I've removed these settings here and have them default to `false`, with some code in the browser initialisation to set the to `true`.

lib/webenginesettings.cpp
llewelld: I've moved this out of here too.

llewelld:
The most recent commit https://github.com/sailfishos/sailfish-components-webview/pull/169/commits/4c2429e24deaf136ed35a955bfd95248b204379f needs to be squashed; I've left it separate for now to make the situation described in the discussion around https://github.com/sailfishos/gecko-dev/pull/162#issuecomment-2359261658 clearer.

PR #170, opened 2024-10-02
[components-webview] Create screen resolution to layout.css.devPixelsPerPx mapping. JB#62599

rainemak:
This is rather simple way to map these. At the same, there ain't that many screen resolution that we'd like to map. Thus, this approach is reasoble for our needs.

For testing this, if you have ran browser at least once, you need to remove __PREFS_WRITTEN__ flag file from your device.

> rm $HOME/.local/share/org.sailfishos/browser/__PREFS_WRITTEN__


rainemak:
This could theoretically got to master but I don't see us doing releases from there. So, this should be as good.

rainemak:
For reference this 1080x2520 => 2.608695652173913 has been taken from the same device running Android FF.

lib/webenginesettings.cpp
```diff
@@ -191,20 +204,27 @@ void SailfishOS::WebEngineSettings::initialize()
     qreal touchStartTolerance = dragThreshold / QGuiApplication::primaryScreen()->physicalDotsPerInch();
     engineSettings->setPreference(QString("apz.touch_start_tolerance"), QString("%1f").arg(touchStartTolerance));
 
-    qreal pixelRatio = SAILFISH_WEBENGINE_DEFAULT_PIXEL_RATIO * silicaTheme->pixelRatio();
-    // Round to nearest even rounding factor
-    pixelRatio = qRound(pixelRatio / 0.5) * 0.5;
+    QSize screenResolution = QGuiApplication::primaryScreen()->size();
+    int screenWidth = screenResolution.width();
+
+    qreal pixelRatio = mappedPixelRatio(screenResolution);
```
pvuorela, line 210: Uh, this mapping of exact resolutions feels a bit hacky approach. I can understand doing such for just getting stuff done quickly out of the way, but in that case I could skip building infrastructure for doing more overrides and just have an extra if branch here. Maybe with some TODO type of comment to think twice before adding more.

lib/webenginesettings.cpp
```diff
@@ -191,20 +204,27 @@ void SailfishOS::WebEngineSettings::initialize()
...
+
+    // Try to calculate pixelRatio only if not mapped.
+    if (pixelRatio == -1.0) {
+        pixelRatio = SAILFISH_WEBENGINE_DEFAULT_PIXEL_RATIO * silicaTheme->pixelRatio();
+        // Round to nearest even rounding factor
+        pixelRatio = qRound(pixelRatio / 0.5) * 0.5;
```
pvuorela, line 216: Side note: I don't know the requirements for a good web pixel ratio, but from the overridden value with lots of decimal I wonder if the below calculation is generally good. If there's no need to get "round" ratios (1.5, 2, 3, etc) then maybe silica ratio shouldn't be used for calculation source?

lib/webenginesettings.cpp
rainemak: I'll add a todo. I actually think it other way around. Mapping should be primary and only fallback to something else. This is precise and accurate value rendering good size web content that is well readable whilst not rendering fisher pricy feeling.

pvuorela: But it shouldn't be that surprising encountering devices with exactly the same resolution while quite different screen sizes? Mappings stop working at that point?

pvuorela:
Said resolution being in xperia 10 series they appear to have ppi of 457 (10V 449, not that far). 
-> 457 / 2.608695652173913 = 175.183

If the non-rounded css pixel ratios are ok, pondering should we just on all(?) the devices compare the device ppi to 175.

PR #174, opened 2024-11-04
[sailfish-webview] Update plugins.qmltypes. JB#61769

martyone:


import/webengine/plugins.qmltypes
```diff
@@ -62,6 +62,13 @@ Module {
             name: "setViewCreator"
             Parameter { name: "viewCreator"; type: "QMozViewCreator"; isPointer: true }
         }
+        Method {
+            name: "createView"
+            type: "uint"
+            Parameter { name: "parentId"; type: "uint" }
+            Parameter { name: "parentBrowsingContext"; type: "uintptr_t" }
```
pvuorela, line 69: Not really qml material with uintptr_t but the whole thing here shouldn't really belong here. Not exported either, though, and some other uintptr_t stuff already here. So maybe just good enough.

pvuorela (approved):

PR #178, opened 2025-03-18
Remove some unused properties

pvuorela:
Beef:

    Usage removed with commit 22b6110ce414. I'm assuming this is private
    enough item so we can just remove the properties. The component here
    isn't referenced in the module documentation and the earlier usage
    seemed like internal handling.


direc85 (approved):
Nothing wrong with removing unused and undocumented properties, I think. And even if someone had used them, it would have been ineffective/broken since 2021 already.

PR #183, opened 2026-05-22
Adapt Sailfish WebView for ESR115

abranson:
- Adapt WebView settings and renderer preferences for the ESR115 stack.
- Forward safe-area insets and handle unavailable Gecko scroll metrics.
- Add Sailfish color picker handling and hide unavailable auth remember controls.
- Apply Sailfish text size settings to WebView content.
- Stop WebView embedding before application quit to avoid Gecko shutdown crashes.

### DIFF
--- a/import/pickers/WebColorPickerPage.qml
+++ b/import/pickers/WebColorPickerPage.qml
@@ -1,0 +1,79 @@
+/****************************************************************************
+**
+** Copyright (c) 2026 Jolla Mobile Ltd
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.6
+import Sailfish.Silica 1.0
+
+ColorPickerPage {
+    id: colorPickerPage
+
+    property int winId
+    property QtObject contentItem
+    property string initialColor: "#000000"
+    property var defaultColors: []
+    property bool _completed
+
+    color: _cssColor(initialColor) || "#000000"
+
+    function _cssColor(value) {
+        var color = String(value).toLowerCase()
+        if (color.length === 9 && color.charAt(0) === "#") {
+            color = "#" + color.substring(3)
+        }
+        if (color.length !== 7 || color.charAt(0) !== "#") {
+            return ""
+        }
+
+        for (var i = 1; i < color.length; ++i) {
+            if ("0123456789abcdef".indexOf(color.charAt(i)) < 0) {
+                return ""
+            }
+        }
+        return color
+    }
+
+    function _sendResponse(accepted, selectedColor) {
+        if (contentItem) {
+            contentItem.sendAsyncMessage("embedui:colorpickerresponse",
+                                         {
+                                             "winId": winId,
+                                             "accepted": accepted,
+                                             "color": accepted ? selectedColor : ""
+                                         })
+        }
+    }
+
+    Component.onCompleted: {
+        if (defaultColors && defaultColors.length > 0) {
+            colors = defaultColors
+        }
+    }
+
+    Component.onDestruction: {
+        if (!_completed) {
+            _completed = true
+            _sendResponse(false, "")
+        }
+    }
+
+    onColorClicked: {
+        var selectedColor = _cssColor(color)
+        _completed = true
+        _sendResponse(selectedColor.length > 0, selectedColor)
+        pageStack.pop()
+    }
+
+    on_NavigationChanged: {
+        if (_navigation == PageNavigation.Back && !_completed) {
+            _completed = true
+            _sendResponse(false, "")
+        }
+    }
+}
--- a/import/webview/TextZoomController.qml
+++ b/import/webview/TextZoomController.qml
@@ -1,0 +1,54 @@
+/****************************************************************************
+**
+** Copyright (c) 2026 Jolla Mobile Ltd
+**
+****************************************************************************/
+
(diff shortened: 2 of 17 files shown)


import/pickers/WebColorPickerPage.qml
```diff
@@ -0,0 +1,79 @@
...
+**
+** Copyright (c) 2026 Jolla Ltd.
+**
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
```
pvuorela: New files could do 2.6 by now. Applies to some others too.

import/webview/TextZoomController.qml
```diff
@@ -0,0 +1,70 @@
...
+    property var webPage
+    property bool _frameScriptLoaded
+
+    property ConfigurationValue fontSizeCategoryConfig: ConfigurationValue {
+        key: "/desktop/jolla/theme/font/sizeCategory"
+        defaultValue: "normal"
+    }
+
+    readonly property real textZoom: {
+        switch (fontSizeCategoryConfig.value) {
```
pvuorela: Alternatively could consider just using something like Theme.fontSizeMedium / Theme.fontSizeMediumBase. Think it should give the same result without needing to get into lower level dconf or duplicating the zoom levels.

import/webview/plugin.cpp
```diff
@@ -28,11 +34,246 @@
...
+            }
+            if (m_waitLoop) {
+                m_waitLoop->quit();
+            }
+            if (m_quitAfterShutdown) {
+                restoreQuitOnLastWindowClosed();
+                QCoreApplication::quit();
+            }
+        });
+        if (QGuiApplication *app = qobject_cast<QGuiApplication *>(QCoreApplication::instance())) {
```
pvuorela: Optionally qGuiApp which handles the casting.

import/pickers/WebColorPickerPage.qml
rainemak: Irrelevant.

import/pickers/WebColorPickerPage.qml
```diff
@@ -0,0 +1,79 @@
+/****************************************************************************
+**
+** Copyright (c) 2026 Jolla Ltd.
```
rainemak: Jolla Mobile Ltd without dot at the end.

import/pickers/WebColorPickerPage.qml
```diff
@@ -0,0 +1,79 @@
...
+****************************************************************************/
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import QtQuick 2.0
+import Sailfish.Silica 1.0
+
+ColorPickerPage {
```
rainemak, line 14: Fyi, close also old color picker task JB#52212

import/pickers/WebColorPickerPage.qml
```diff
@@ -0,0 +1,79 @@
...
+            if ("0123456789abcdef".indexOf(color.charAt(i)) < 0) {
+                return ""
+            }
+        }
+        return color
+    }
+
+    function _sendResponse(accepted, selectedColor) {
+        if (contentItem) {
+            contentItem.sendAsyncMessage("colorpickerresponse",
```
rainemak: Is the engine side message really like this? I may stand corrected regarding the previous "embed:" & "embedui:" comment.

import/webview/TextZoomController.qml
rainemak: Does the math really work?

import/webview/TextZoomController.qml
```diff
@@ -0,0 +1,70 @@
...
+
+    function updateTextZoom() {
+        if (!webPage) {
+            return
+        }
+
+        if (!_frameScriptLoaded) {
+            webPage.loadFrameScript(Qt.resolvedUrl("TextZoom.js"))
+            _frameScriptLoaded = true
+        }
```
rainemak, line 39: Why don't we load this in all scenarios? I see this conditional loading as waste. Could you elaborate?

rpm/sailfish-components-webview.spec
```diff
@@ -1,6 +1,8 @@
+%global min_qtmozembed_version 1.56.0
+
 Name:    sailfish-components-webview-qt5
 Summary: Allows embedding Sailfish WebView into applications
-Version: 1.6.0
+Version: 1.7.0
```
rainemak, line 5: Directly 2.0.0

rainemak (changes requested):
No major comments but please check my comments though.

import/pickers/WebColorPickerPage.qml
pvuorela: Sure, not mandatory. Just trying to push a bit of a convention with these so people don't waste time wondering why things like Column bottomPadding/topPadding don't work.

import/webview/TextZoomController.qml
pvuorela: Sure it should, fontsizeMedium is base size * multiplier.

import/webview/TextZoomController.qml
abranson: This is to make sure it only gets loaded once, rather than in different scenarios.

rainemak (approved):

rpm/sailfish-components-webview.spec
abranson: This caused an so version bump. Not sure if that's a good idea.

PR #184, opened 2026-06-10
Fix webview orientation updates.

abranson:
Request a scene graph update after content orientation messages so the reoriented frame is presented without waiting for later interaction.

Also stop flipping gesture deltas for inverted portrait, where item coordinates already match the presented orientation.

### DIFF
--- a/import/webview/rawwebview.cpp
+++ b/import/webview/rawwebview.cpp
@@ -259,14 +259,15 @@ void RawWebView::touchEvent(QTouchEvent *event)
                     QPointF delta = touchPoint.scenePos() - m_startPos;
 
                     switch (orientation()) {
+                    case Qt::PortraitOrientation:
+                    case Qt::InvertedPortraitOrientation:
+                        // Item coordinates already match the presented orientation.
+                        break;
                     case Qt::LandscapeOrientation:
                         delta = QPointF(delta.y(), -delta.x());
                         break;
                     case Qt::InvertedLandscapeOrientation:
                         delta = QPointF(-delta.y(), delta.x());
-                        break;
-                    case Qt::InvertedPortraitOrientation:
-                        delta = QPointF(-delta.x(), -delta.y());
                         break;
                     default:
                         break;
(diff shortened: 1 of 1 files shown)


import/webview/rawwebview.cpp
```diff
@@ -322,16 +319,19 @@ void RawWebView::touchEvent(QTouchEvent *event)
...
-        } else if (orientation == QStringLiteral("portrait-secondary")) {
+        } else if (orientationName == QStringLiteral("portrait-secondary")) {
             mappedOrientation = Qt::InvertedPortraitOrientation;
         }
         emit contentOrientationChanged(mappedOrientation);
+        // Force a fresh scene-graph update so the reoriented WebRender frame
+        // is presented without waiting for additional user interaction.
+        update();
     }
 }
```
direc85, line 345: I'd love if this `if` would cover the fourth case as well, something like this:

```c++
void RawWebView::onAsyncMessage(const QString &message, const QVariant &data)
{
    if (message == CONTENT_ORIENTATION_CHANGED) {
        const QString orientationName = data.toMap().value("orientation").toString();
        if (orientationName == QStringLiteral("portrait-primary")) { // correct name?
            emit contentOrientationChanged(Qt::PortraitOrientation);
        } else if (orientationName == QStringLiteral("landscape-primary")) {
            emit contentOrientationChanged(Qt::LandscapeOrientation);
        } else if (orientationName == QStringLiteral("landscape-secondary")) {
            emit contentOrientationChanged(Qt::InvertedLandscapeOrientation);
        } else if (orientationName == QStringLiteral("portrait-secondary")) {
            emit contentOrientationChanged(Qt::InvertedPortraitOrientation);
        } else {
            // log?
        }
        // Force a fresh scene-graph update so the reoriented WebRender frame
        // is presented without waiting for additional user interaction.
        update();
    }
}
```

import/webview/rawwebview.cpp
```diff
@@ -265,9 +265,6 @@ void RawWebView::touchEvent(QTouchEvent *event)
                     case Qt::InvertedLandscapeOrientation:
                         delta = QPointF(-delta.y(), delta.x());
                         break;
-                    case Qt::InvertedPortraitOrientation:
-                        delta = QPointF(-delta.x(), -delta.y());
-                        break;
```
direc85, line 270: Could maybe add the two unneeded cases as fallthroughs for completeness's sake, maybe with a comment.

direc85 (approved):
A few suggestions, but nothing blocking. Seems to work well!

import/webview/rawwebview.cpp
abranson: Much neater. I initialized it as PrimaryOrientation as that seemed like a better fallback. There's no longer in here at all yet though so I thought it was a bit much to bring it in just for this.

import/webview/rawwebview.cpp
abranson: Added it as another case for the Portrait condition

